PR #185, opened 2025-12-10
fix: Support string key parameters in Lisp runtime functions

andreasronge:
## Summary

- Fixed `FunctionClauseError` when LLMs generate string keys (`"price"`) instead of keywords (`:price`) in PTC-Lisp programs
- Extended `flex_get` helper to support bidirectional key matching (atom↔string)
- Added `mix lisp` alias to demo for easier CLI access

## Changes

**Runtime fixes (`lib/ptc_runner/lisp/runtime.ex`):**
- Extended `flex_get/2` to handle string keys by trying `String.to_existing_atom/1` conversion
- Updated `sort_by/2` and `sort_by/3` guards to accept both atoms and strings
- Changed `get/2` and `get/3` to use `flex_get` for bidirectional key matching
- Updated `contains?/2` for maps to check both atom and string versions of keys

**Tests (`test/ptc_runner/lisp/runtime_test.exs`):**
- Added 20 new tests covering string key parameters for: `sort_by`, `sum_by`, `avg_by`, `min_by`, `max_by`, `group_by`, `pluck`, `get`, `contains?`

**Documentation (`docs/ptc-lisp-specification.md`):**
- Updated "Flexible Key Access" section to describe bidirectional key matching
- Added examples showing string key parameters work with LLM-generated code

**Demo (`demo/mix.exs`):**
- Added `mix lisp` alias as documented in README

## Test plan

- [x] All 1034 existing tests pass
- [x] 20 new string key parameter tests pass
- [x] Code formatting verified
- [x] No compiler warnings

### DIFF
--- a/lib/ptc_runner/lisp/runtime.ex
+++ b/lib/ptc_runner/lisp/runtime.ex
@@ -170,9 +170,23 @@ defmodule PtcRunner.Lisp.Runtime do
   def get(nil, _k), do: nil
 
   def get(m, k, default) when is_map(m) do
-    case flex_get(m, k) do
-      nil -> default
-      val -> val
+    cond do
+      Map.has_key?(m, k) ->
+        Map.get(m, k)
+
+      is_atom(k) and Map.has_key?(m, to_string(k)) ->
+        Map.get(m, to_string(k))
+
+      is_binary(k) ->
+        try do
+          atom_key = String.to_existing_atom(k)
+          if Map.has_key?(m, atom_key), do: Map.get(m, atom_key), else: default
+        rescue
+          ArgumentError -> default
+        end
+
+      true ->
+        default
     end
   end
 
--- a/test/ptc_runner/lisp/runtime_test.exs
+++ b/test/ptc_runner/lisp/runtime_test.exs
@@ -130,6 +130,18 @@ defmodule PtcRunner.Lisp.RuntimeTest do
       map = %{"name" => "Alice"}
       assert Runtime.get(map, "age", 0) == 0
     end
+
+    test "get with default preserves explicit nil values" do
+      map = %{status: nil}
+      assert Runtime.get(map, :status, "unknown") == nil
+      assert Runtime.get(map, "status", "unknown") == nil
+    end
+
+    test "get with default and string key preserves explicit nil values in string-keyed map" do
+      map = %{"status" => nil}
+      assert Runtime.get(map, "status", "unknown") == nil
+      assert Runtime.get(map, :status, "unknown") == nil
+    end
   end
 
   describe "string key parameters - contains?" do


andreasronge:
@claude please fix these issues from PR review:

## Issue 1: Bug - get/3 returns default for explicit nil values
**Location**: `lib/ptc_runner/lisp/runtime.ex:172-177`
**Problem**: The current `get/3` implementation cannot distinguish between "key not found" and "key exists with nil value". Both cases return the default value, which is a semantic regression from standard `Map.get/3` behavior.
**Fix**: Update the `get/3` implementation to use `Map.has_key?/2` or `Map.fetch/1` to check if the key exists before falling back to the default. The logic should be:
1. First check if the key exists in the map (trying both the provided key and its atom/string variant via flex_get logic)
2. If the key exists, return the value (even if it's nil)
3. Only return the default if the key doesn't exist at all

Example fix approach:
```elixir
def get(m, k, default) when is_map(m) do
  cond do
    Map.has_key?(m, k) -> Map.get(m, k)
    # Try string/atom variant
    is_atom(k) and Map.has_key?(m, to_string(k)) -> Map.get(m, to_string(k))
    is_binary(k) -> 
      try do
        atom_key = String.to_existing_atom(k)
        if Map.has_key?(m, atom_key), do: Map.get(m, atom_key), else: default
      rescue
        ArgumentError -> default
      end
    true -> default
  end
end
```

## Issue 2: Missing test for nil value edge case
**Location**: `test/ptc_runner/lisp/runtime_test.exs`
**Problem**: No test verifies that `get/3` correctly returns nil when a key exists with an explicit nil value.
**Fix**: Add test cases to the "string key parameters - get" describe block:
```elixir
test "get with default preserves explicit nil values" do
  map = %{status: nil}
  assert Runtime.get(map, :status, "unknown") == nil
  assert Runtime.get(map, "status", "unknown") == nil
end

test "get with default and string key preserves explicit nil values in string-keyed map" do
  map = %{"status" => nil}
  assert Runtime.get(map, "status", "unknown") == nil
  assert Runtime.get(map, :status, "unknown") == nil
end
```

Please fix ALL issues above in a SINGLE commit. Make minimal changes to resolve each issue.

andreasronge:
## Fix Applied

Both issues from the PR review have been fixed in a single commit:

### Issue 1: Fixed get/3 nil value handling
Updated `get/3` in `lib/ptc_runner/lisp/runtime.ex` to use `Map.has_key?/2` checks to distinguish between "key not found" and "key exists with nil value". The implementation now:
1. Checks if the key exists in the map directly
2. Tries the atom/string variant if the key is an atom
3. Safely tries to convert string keys to atoms using `String.to_existing_atom/1`
4. Returns the default only if the key truly doesn't exist

This restores standard `Map.get/3` behavior where `get(%{status: nil}, :status, "unknown")` correctly returns `nil` instead of `"unknown"`.

### Issue 2: Added nil edge case tests
Added two test cases to `test/ptc_runner/lisp/runtime_test.exs` in the "string key parameters - get" describe block:
- `get with default preserves explicit nil values` - Tests both atom and string key lookups
- `get with default and string key preserves explicit nil values in string-keyed map` - Tests with string-keyed maps

### Verification
- All 1036 tests pass
- Credo (code style) checks pass
- Code formatting verified
- Pre-commit checks passed

andreasronge:
## Fix Applied

Both issues from the PR review have been fixed in a single commit:

### Issue 1: Fixed get/3 nil value handling
Updated `get/3` in `lib/ptc_runner/lisp/runtime.ex` to use `Map.has_key?/2` checks to distinguish between "key not found" and "key exists with nil value". The implementation now:
1. Checks if the key exists in the map directly
2. Tries the atom/string variant if the key is an atom
3. Safely tries to convert string keys to atoms using `String.to_existing_atom/1`
4. Returns the default only if the key truly doesn't exist

This restores standard `Map.get/3` behavior where `get(%{status: nil}, :status, "unknown")` correctly returns `nil` instead of `"unknown"`.

### Issue 2: Added nil edge case tests
Added two test cases to `test/ptc_runner/lisp/runtime_test.exs` in the "string key parameters - get" describe block:
- `get with default preserves explicit nil values` - Tests both atom and string key lookups
- `get with default and string key preserves explicit nil values in string-keyed map` - Tests with string-keyed maps

### Verification
- All 1036 tests pass
- Credo (code style) checks pass
- Code formatting verified
- Pre-commit checks passed

Issue #220, opened 2025-12-11
Add memory support to JSON Agent


andreasronge:
## Summary

Add memory support to the JSON `Agent` to enable multi-turn tests where values can be stored across queries. This mirrors the `LispAgent` implementation and allows the `JsonTestRunner` to run multi-turn test cases.

## Context

**Architecture reference**: [Test Runner Refactoring Plan - Phase 4](https://github.com/andreasronge/ptc_runner/blob/main/demo/docs/test-runner-refactor-plan.md#phase-4-multi-turn-support-enhancement) - Task 4.1
**Dependencies**: #214 (JsonTestRunner created) - completed
**Related issues**: #195 (Epic: Unify Demo Test Runners)

## Current State

- `PtcDemo.Agent` (574 lines) does not have memory support
- `PtcDemo.LispAgent` already has memory support via `:memory` field in struct (line 35)
- `PtcRunner.Json` supports loading values from context via `{"op": "load", "name": "X"}`
- The spec document resolved that JSON DSL does NOT support persistent memory operations, so we implement agent-level memory (Option B from spec)

**Key difference from Lisp**: `PtcRunner.Lisp.run/2` returns `{:ok, result, _delta, new_memory}` with built-in memory persistence. For JSON, we need to implement memory at the Agent level by:
1. Detecting "store as {name}" patterns in queries
2. Storing results in Agent state
3. Injecting stored values into context with `memory_` prefix

## Acceptance Criteria

- [ ] `Agent` struct includes `:memory` field (map of name -> value)
- [ ] `agent_loop/6` becomes `agent_loop/7` accepting and returning memory parameter
- [ ] When query contains "store as {name}" pattern, the result is stored in memory
- [ ] Memory values are injected into context with `memory_` prefix for program execution
- [ ] Memory persists across queries within a session
- [ ] `Agent.reset/0` clears memory
- [ ] `Agent.set_data_mode/1` clears memory
- [ ] System prompt documents memory convention for JSON DSL
- [ ] Code compiles: `cd demo && mix compile --warnings-as-errors`
- [ ] Demo tests pass: `cd demo && mix test`

## Implementation Hints

**Files to modify:**
- `demo/lib/ptc_demo/agent.ex` - Add memory support

**Struct change (line 27):**
```elixir
# Current:
defstruct [:model, :context, :datasets, :last_program, :last_result, :data_mode, :usage]

# After:
[...]

PR #221, opened 2025-12-11
feat: Add memory support to JSON Agent (#220)

andreasronge:
## Summary

Implements memory persistence for the JSON Agent, mirroring the LispAgent pattern. This enables multi-turn queries where values can be stored and retrieved across questions using the "store as X" pattern.

## Changes

- **Agent struct**: Added `:memory` field (map of name -> value)
- **Initialization**: Memory initialized as empty map in `init/1`
- **agent_loop/7**: Updated to accept and return memory parameter
- **Memory injection**: Memory values injected into context with `memory_` prefix before program execution
- **Store detection**: Regex pattern detects "store as {name}" in user queries and persists results
- **Memory cleanup**: Reset on `reset()` and `set_data_mode()` calls
- **Documentation**: Updated system prompts (schema and explore modes) and module documentation

## Implementation Details

Memory works by:
1. Injecting stored values into the context as `memory_{key}` variables
2. Making them accessible via `{"op": "load", "name": "memory_{key}"}`
3. Detecting "store as X" patterns (case-insensitive) in user questions
4. Persisting results in the Agent state for subsequent queries

## Test Results

✅ All 125 tests pass
✅ Code formatting verified
✅ Compilation successful with warnings-as-errors
✅ Credo checks passed

Completes Phase 4.1 of the test runner refactoring plan.

### DIFF
--- a/demo/lib/ptc_demo/agent.ex
+++ b/demo/lib/ptc_demo/agent.ex
@@ -332,7 +332,7 @@ defmodule PtcDemo.Agent do
 
                 # Detect "store as {name}" pattern in the original query
                 new_memory =
-                  case Regex.run(~r/store (?:it |the result |this )?as (\w+)/i, original_query) do
+                  case Regex.run(~r/store (?:it |the result |this )?as ([\w-]+)/i, original_query) do
                     [_, name] -> Map.put(memory, name, result)
                     nil -> memory
                   end


andreasronge:
@claude please fix these issues from PR review:

## Issue 1: Regex doesn't match hyphenated memory keys
**Location**: `demo/lib/ptc_demo/agent.ex:335`
**Problem**: The regex `~r/store (?:it |the result |this )?as (\w+)/i` uses `\w+` which only matches word characters (alphanumeric + underscore), not hyphens. This means multi-turn test cases that use hyphenated keys like `delivered-count` and `engineering-employees` will fail to store values in memory.
**Fix**: Change the regex from `(\w+)` to `([\w-]+)` to include hyphens in the capture group. This matches the issue specification which explicitly states: "Store pattern with hyphens: 'store as delivered-count' should store under key 'delivered-count'"

Please fix this issue in a SINGLE commit. Make minimal changes to resolve the issue.

andreasronge:
## Fix Applied

**Issue**: Regex doesn't match hyphenated memory keys

**Solution**: Updated the regex pattern in `demo/lib/ptc_demo/agent.ex:335` from `(\w+)` to `([\w-]+)` to capture hyphenated memory keys.

**Changes**:
- Modified regex capture group to support hyphens: `[\w-]` instead of `\w`
- This allows multi-turn test cases to store values under hyphenated keys like `delivered-count` and `engineering-employees`

**Verification**:
- ✅ All tests pass (1081 tests, 0 failures)
- ✅ Code formatting verified
- ✅ Compilation with warnings-as-errors successful
- ✅ Credo analysis passed

The fix is minimal and focused, addressing exactly the issue raised in the PR review without any unrelated changes.

PR #250, opened 2025-12-11
fix: Resolve remaining Clojure conformance test failures

### LINKED ISSUE #245: fix: Align PTC-Lisp semantics with Clojure specification
## Summary

Several PTC-Lisp functions have semantic differences from Clojure that should be fixed to ensure programs behave identically when run in both environments. This affects `or`, `some`, and variadic subtraction.

## Context

**Architecture reference**: [PTC-Lisp Specification](docs/ptc-lisp-specification.md), [Guide](docs/guide.md)
**Dependencies**: None
**Related issues**: None

### 1. `or` returns `nil` instead of last falsy value
```clojure
(or false false)
;; PTC-Lisp: nil
;; Clojure:  false

(or nil false)
;; PTC-Lisp: nil
;; Clojure:  false

(or nil nil)
;; PTC-Lisp: nil
;; Clojure:  nil (correct by coincidence)
```

**Location**: `lib/ptc_runner/lisp/eval.ex:288`

**Root cause**: The empty case `do_eval_or([], ...)` returns `nil`, but should return the last evaluated value. Need to track last value through recursion.

### 2. `some` returns boolean instead of first truthy result/nil
```clojure
(some even? [1 3 5])
;; PTC-Lisp: false
;; Clojure:  nil

(some even? [1 2 3])
;; PTC-Lisp: true
;; Clojure:  true (correct)
```

**Location**: `lib/ptc_runner/lisp/runtime.ex:198`

**Root cause**: Currently uses `Enum.any?(coll, pred)` which returns a boolean. Clojure's `some` returns the first truthy predicate result, or `nil` if none.

### 3. Variadic subtraction incorrect for 3+ args AND unary negation broken
```clojure
(- 10)
;; PTC-Lisp: 10
;; Clojure:  -10 (negation)

(- 10 3 2)
;; PTC-Lisp: 9
;; Clojure:  5  (10 - 3 - 2)

(- 10 3 2 1)
;; PTC-Lisp: -8
;; Clojure:  4  (10 - 3 - 2 - 1)
```

**Locations**:
- `lib/ptc_runner/lisp/eval.ex:411-422` (variadic handler)
- `lib/ptc_runner/lisp/eval.ex:442-448` (unary minus - **unreachable code**)

**Root causes**:
1. **Argument order**: `Enum.reduce(t, h, &Kernel.-/2)` passes `(element, accumulator)`, so it computes `element - accumulator` instead of `accumulator - element`
2. **Clause ordering bug**: The unary minus special case at lines 442-448 comes AFTER the general variadic handler at 411-422. Since the general handler matches `[x] -> x` first, the unary minus handler is never reached.

### Known Intentional Differences (Out of Scope)

These are intentional design decisions documented in `docs/ptc-lisp-specification.md` Section 13:
- Sequence functions (`filter`, `map`, `sort`, etc.) return vectors instead of lazy seqs
- This is practical for LLM use cases and doesn't affect program correctness

## Acceptance Criteria

- [ ] `(or false false)` returns `false`
- [ ] `(or nil false)` returns `false` (last value)
- [ ] `(or nil nil)` returns `nil` (last value)
- [ ] `(some even? [1 3 5])` returns `nil`
- [ ] `(some even? [1 2 3])` returns `true` (or truthy value)
- [ ] `(- 10)` returns `-10` (negation)
- [ ] `(- 10 3)` returns `7`
- [ ] `(- 10 3 2)` returns `5`
- [ ] `(- 10 3 2 1)` returns `4`
- [ ] All existing tests pass

### Fix 1: `or` - Track last evaluated value

**File**: `lib/ptc_runner/lisp/eval.ex`

Change `do_eval_or/5` to pass the last evaluated value instead of returning `nil` for empty list:
```elixir
# Base case: return last evaluated value (nil if never evaluated)
defp do_eval_or([], last_value, _ctx, memory, _env, _tool_exec),
  do: {:ok, last_value, memory}

# Recursive case: track last value
defp do_eval_or([e | rest], _last, ctx, memory, env, tool_exec) do
  with {:ok, value, memory2} <- do_eval(e, ctx, memory, env, tool_exec) do
    if truthy?(value) do
      {:ok, value, memory2}
    else
      do_eval_or(rest, value, ctx, memory2, env, tool_exec)
    end
  end
end
```

**Pattern to follow**: See `do_eval_and/5` for similar short-circuit pattern (though `and` returns first falsy, not last).

### Fix 2: `some` - Return predicate result or nil

**File**: `lib/ptc_runner/lisp/runtime.ex:198`

Replace the `Enum.any?` call with a find-first-truthy pattern:
```elixir
def some(pred, coll) when is_list(coll) do
  Enum.find_value(coll, fn x ->
    result = pred.(x)
    if result, do: result, else: nil
  end)
end
```

### Fix 3: Subtraction - Fix reduce order and clause ordering

**File**: `lib/ptc_runner/lisp/eval.ex`

Two changes needed:

1. **Move unary minus before general variadic handler** (around line 411):
```elixir
# Special handling for unary minus - MUST come before general variadic
defp apply_fun({:variadic, fun2, _identity}, [x], _ctx, memory, _tool_exec)
     when fun2 == (&Kernel.-/2) do
  {:ok, -x, memory}
end

# General variadic handler
defp apply_fun({:variadic, fun2, identity}, args, ...) do
  # ...existing code but fix reduce for subtraction...
end
```

2. **Fix the reduce logic for subtraction** (keep accumulator on left):
```elixir
[h | t] ->
  if fun2 == (&Kernel.-/2) do
    Enum.reduce(t, h, fn elem, acc -> acc - elem end)
  else
    Enum.reduce(t, h, fun2)
  end
```

**Alternative approach**: Change subtraction binding in `env.ex` to use `{:multi_arity, ...}` with separate implementations for 1-arg, 2-arg, and n-arg cases. This may be cleaner.

## Test Plan

**Unit tests to add/verify** (in existing test files):

1. `test/ptc_runner/lisp/eval_test.exs` - `or` with various combinations of `nil`, `false`, truthy values
2. `test/ptc_runner/lisp/runtime_test.exs` - `some` with predicates that match and don't match
3. `test/ptc_runner/lisp/integration_test.exs` - `-` with 1, 2, 3, and 4+ arguments

**Edge cases**:
- `(or)` with no args should return `nil`
- `(- x)` with single arg should return `-x` (negation)
- `(some pred [])` with empty collection should return `nil`
- `(some nil? [1 nil 3])` should return `true` (predicate returns `true` for `nil`)

**Verification command**:
```bash
mix test
```

## Out of Scope

- Changing sequence functions to return lazy seqs (intentional difference per Section 13.1)
- Adding Clojure conformance test suite (separate issue if desired)
- Other Clojure features not in PTC-Lisp spec

andreasronge:
## Summary

Fix multiple issues causing 15 Clojure conformance tests to fail after the initial Clojure validation implementation.

### ClojureValidator EDN parsing (fixed 12 test failures)
- Add support for parsing Clojure lists `(1 2 3)` in `parse_simple_edn`
- Use `read-string` to safely parse EDN output before JSON conversion in `parse_edn_output`
- This prevents lists like `(2 4)` from being interpreted as function calls when converting to JSON

### Variadic multiplication `(* 5)` (fixed 1 failure)
- Fix single-argument handler in `apply_fun` to return the value for all variadic ops
- Previously only `-` (negation) was handled correctly, others returned `{:error, {:not_callable, ...}}`

### Variadic division `(/ 100 5 2)` (fixed 1 failure)
- Change division from `:normal` (binary only) to `:variadic_nonempty` in env.ex
- Now correctly reduces left-to-right: `(/ 100 5 2)` = 100 / 5 / 2 = 10

### 2-arity reduce `(reduce + [1 2 3])` (fixed 1 failure)
- Add `reduce/2` in runtime.ex that uses first element as initial value
- Change binding to `:multi_arity` to support both 2 and 3 argument forms

## Test plan

All 63 Clojure conformance tests now pass:

```bash
mix test --only clojure
# 63 tests, 0 failures
```

Full test suite also passes:

```bash
mix test --exclude clojure --exclude e2e
# 1126 tests, 0 failures
```

Closes #245


andreasronge:
## Auto-Triage Summary (Cycle 2)

### Review Analysis
Analyzed two review comments from github-actions bot. Both reviews provide **APPROVE** verdicts with only optional suggestions.

### Decisions Made
| # | Issue | Decision | Action |
|---|-------|----------|--------|
| 1 | File size - test files exceed 800 line threshold | ALREADY_TRACKED | Issue #251 exists |
| 2 | Demo report files gitignore consideration | IGNORE | Low priority, speculative concern |

### Rationale

**Issue #1 - File size (ALREADY_TRACKED)**
- Previous triage correctly created issue #251
- Both review comments acknowledge this is tracked
- No additional action needed

**Issue #2 - Demo reports (IGNORE)**
- Framed as "consider whether" (speculative)
- Previous triage correctly determined these provide documentation value
- No concrete problem identified
- Consistent with first triage decision

### Status
- FIX_NOW items: 0
- New issues created: 0 (existing issue #251 covers file size concern)
- Items ignored: 1
- **PR remains ready-to-merge** ✓

**Conclusion**: The previous triage handled all review suggestions appropriately. No additional fixes or issues needed. The PR is ready for merge once all checks pass.

PR #897, opened 2026-05-11
perf(precommit): faster precommit — drop forced recompile, split out dialyzer, parallelize spec validation

andreasronge:
## Summary

Investigation into `mix precommit` performance, plus the low-risk wins it surfaced.

**Where the time went** (warm run ≈ 50s wall on a heavily-loaded machine):
| Step | Cost | Note |
|---|---|---|
| `compile --force --warnings-as-errors` | ~5–8s | full recompile of all 152 `.ex` files **every run**, even with zero changes |
| `credo --strict` | ~5s | fine |
| `dialyzer` | ~5s warm / **2–3 min on PLT rebuild** | the lurking cost whenever a dep changes |
| `ptc.validate_spec` | ~15s | 380 spec examples run serially |
| `test --warnings-as-errors` | ~15–20s | 4831 tests, already well-parallelized |
| demo / ptc_viewer tests | ~2s warm | (separate `_build`, cheap when warm) |

The alias already runs in a single BEAM process, so per-step VM-startup overhead was not a factor.

## Changes

- **Drop `--force` from the compile step.** Eliminates the ~5–8s full recompile on every run. `mix compile --warnings-as-errors` still surfaces warnings on changed files; CI's clean build remains the authoritative warnings-as-errors gate.
- **Move `dialyzer` (+ `deps.unlock --check-unused`) out of `precommit` into a new `mix prepush` alias.** Dialyzer is cheap when the PLT is warm but costs 2–3 min on a rebuild; CI runs it on every PR regardless. `precommit` stays the fast per-commit gate; `prepush` is the heavier pre-push check.
- **Parallelize `ptc.validate_spec`** with `Task.async_stream` (each example already runs in its own sandbox process). ~15s → ~3s standalone. Ordering of `failures`/`by_section` is preserved (`ordered: true`); task crashes/timeouts are now reported as failures rather than aborting the run.
- **Add an empty `.dialyzer_ignore.exs` + `list_unused_filters: true`** to silence dialyxir's "No :ignore_warnings opt specified" notice and prevent stale filters accumulating.
- **Update `CLAUDE.md`** to document the `precommit` / `prepush` split.

Net: `mix precommit` dropped from ~52s to ~32s in back-to-back runs here (both noisy — machine was at load ~50 on 10 cores), and no longer does a redundant full recompile.

## Considered but not done

- **Parallel runner for the whole alias** — concurrent `mix` processes contend on Mix's build lock during their compile-check phase, so the win is modest and the failure modes are flaky. Left as a sequential alias.
- **`schema.gen` as a check instead of regenerate** — it's cheap inside the shared BEAM; not worth the churn.

## Verification

- `mix precommit` → green (format, compile, credo, schema.gen, ptc.validate_spec 380/380, 4831 tests, demo 200 tests, ptc_viewer 9 tests)
- `mix prepush` → green (dialyzer passed, unused-deps check passed)

## Fix Automation State

Fix attempts: 2/3

### DIFF
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -1,0 +1,4 @@
+# Dialyzer warning filters. Keep empty unless a warning is a confirmed false
+# positive — each entry is either a regex string or a {file, warning} tuple.
+# `list_unused_filters: true` (mix.exs) flags entries here that no longer match.
+[]
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,12 +30,13 @@ mix test                        # Run tests (--failed to re-run failures)
 mix format                      # Format code
 mix credo --strict              # Lint
 mix dialyzer                    # Type checking
-mix precommit                   # Run before pushing (format + compile + credo + dialyzer + test)
+mix precommit                   # Fast quality gate (format + compile + credo + schema + spec + tests)
+mix prepush                     # Slower checks before pushing (dialyzer + unused-deps check)
 ```
 
 Full quality check: `mix format --check-formatted && mix compile --warnings-as-errors && mix test`
 
-Always run `mix precommit` before `git push`. If it fails, fix all issues before pushing.
+Run `mix precommit` before committing and `mix prepush` before `git push`. If either fails, fix all issues before pushing. (CI also runs dialyzer and the unused-deps check on every PR.)
 
 ## Project Structure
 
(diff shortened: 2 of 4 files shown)


andreasronge:
## Fix Summary

✅ **Issue Fixed:** Missing `on_timeout: :kill_task` in `Task.async_stream` options in `spec_validator.ex`

Added `on_timeout: :kill_task` to the `Task.async_stream` call at line 882. Without this, a task hitting the 30s timeout would cause the caller process to exit (OTP default `on_timeout: :exit` behaviour). With this fix, timed-out tasks are killed and return `{:exit, :timeout}` tuples, which are correctly handled by the existing `fold_example_result({:exit, reason}, results)` clause — reporting them as failures rather than crashing the entire validation run.

PR #933, opened 2026-05-13
fix(validator): close alias-key bypass in closed_map + add closed_map to JsonHandler

### LINKED ISSUE #932: Validator: alias-key bypass of additionalProperties:false + {:closed_map} not handled in JsonHandler.atomize_value
Follow-up review on PR #906 (validator + MCP `additionalProperties: false` / `{:closed_map, _}`) surfaced two related correctness gaps in `lib/ptc_runner/sub_agent/`. Both have working repros and are not covered by the existing test suite.

## P1 — `additionalProperties: false` is bypassable via alias keys

`PtcRunner.SubAgent.Signature.Validator.validate/2` accepts a map carrying both the canonical underscore key (typed correctly) **and** a hyphenated alias key (typed *in*correctly), against a schema that should reject the alias'''s value.

Repro:

```elixir
alias PtcRunner.SubAgent.Signature.Validator

data   = %{"user_id" => 1, "user-id" => "bad"}
schema = {:closed_map, [{"user_id", :int}]}

Validator.validate(data, schema)
#=> :ok      ← should be {:error, [...]}

# Sanity: the alias *alone* fails as expected
Validator.validate(%{"user-id" => "bad"}, schema)
#=> {:error, [%{message: "expected int, got string", path: ["user_id"]}]}
```

Same shape reproduced through the MCP surface (`ptc_lisp_execute` with an `output_schema` requiring integer `user_id` and `additionalProperties: false`): a program returning both `:user_id 1` and `:user-id "bad"` yields `isError: false` with `validated %{"user_id" => 1, "user-id" => "bad"}`.

### Mechanism

- `validate_map_fields/3` (`validator.ex:142`) iterates declared fields, picks the *first* candidate `get_field/2` returns, type-checks that value, and stops. The underscore key wins, so the hyphen value is never typed.
- `validate_no_extra_fields/3` (`validator.ex:163`) normalizes every data key through `KeyNormalizer.normalize_key/1` before comparing against declared names. Both `"user_id"` and `"user-id"` collapse to `"user_id"`, so neither is flagged as undeclared.

Net effect: the `"user-id" => "bad"` slot is neither type-checked nor reported as extra. This is a hole in the exact property the closed-map type is meant to enforce.

### Suggested fix

Track which data key was consumed by each declared field in `validate_map_fields`, then have `validate_no_extra_fields` flag every *unconsumed* data key (rather than re-normalizing independently). That closes the bypass and also eliminates the silent two-aliases-one-slot ambiguity.

## P2 — `{:closed_map, fields}` is not integrated into all type-driven consumers

`{:closed_map, _}` has dedicated clauses in:

- `lib/ptc_runner/sub_agent/signature.ex` (`type_to_json_schema/1` at line 430)
- `lib/ptc_runner/sub_agent/signature/renderer.ex` (line 123)
- `lib/ptc_runner/sub_agent/signature/coercion.ex` (line 281)
- `lib/ptc_runner/sub_agent/signature/validator.ex` (line 130)

It does **not** have one in `lib/ptc_runner/sub_agent/loop/json_handler.ex`. `atomize_value/2` has a field-aware `{:map, fields}` clause at line 126, but a closed map falls through to the generic `atomize_value(map, _type)` at line 158, which recurses with `nil` types.

### Consequence

Nested type-driven atomization is lost inside a closed map. Concretely, a `:datetime` field inside `{:closed_map, [{"at", :datetime}]}` does not get the ISO-8601 → `%DateTime{}` coercion that the same field gets inside `{:map, [{"at", :datetime}]}`. Because the validator hard-fails non-`%DateTime{}` on `:datetime` (`validator.ex:90-94`), the closed variant silently rejects inputs the open variant accepts — a semantic divergence between two types pitched as "same thing, stricter on extras."

### Suggested fix

One-line equivalence, mirroring `renderer.ex:123` and `signature.ex:430`:

```elixir
def atomize_value(map, {:closed_map, fields}) when is_map(map),
  do: atomize_value(map, {:map, fields})
```

## Why CI didn'''t catch either

- `validator_test.exs:381-387` only exercises the *single-key* alias case (`%{:"user-id" => 1}` with no underscore sibling), which the current code does handle correctly.
- There is no closed-map × `:datetime` test.

Both fixes should land with regression tests covering exactly these shapes.

## Acceptance criteria

- [ ] **P1**: `Validator.validate(%{"user_id" => 1, "user-id" => "bad"}, {:closed_map, [{"user_id", :int}]})` returns `{:error, [...]}` (alias key flagged as extra or its value type-checked)
- [ ] **P1**: Regression test covering multi-alias-key bypass scenario
- [ ] **P2**: `atomize_value(map, {:closed_map, fields})` delegates to `atomize_value(map, {:map, fields})`
- [ ] **P2**: Regression test: `:datetime` field inside `{:closed_map, ...}` gets ISO-8601 → `%DateTime{}` coercion
- [ ] All existing closed_map and validator tests still pass

## Scope

- P1 is a soundness bug in the feature PR #906 added — treat as P1.
- P2 makes `closed_map` quietly stricter than `map` in a way callers won'''t expect — P1-adjacent.

Original review against PR #906 final head `b9cd5e6` / merged `77129b1`.

## Automation State

| Field | Value |
|-------|-------|
| Status | `SUCCESS` |
| PR | #933 |
| Branch | `claude/932-closed-map-alias-bypass-fix` |
| Attempts | 1 |

**Details:** Both bugs fixed with regression tests. P1: consumed-key tracking in `validate_map_fields_tracking/3` closes the alias-key bypass. P2: one-line `atomize_value` delegation for `{:closed_map, fields}`.

andreasronge:
## Summary
- **P1**: Fix alias-key bypass in `closed_map` validation — when data contained both a canonical key (`"user_id"`) and its hyphenated alias (`"user-id"`), the alias was neither type-checked nor flagged as extra. Now tracks which data key each declared field consumed and flags unconsumed keys directly.
- **P2**: Add missing `{:closed_map, fields}` clause to `JsonHandler.atomize_value/2` — type-driven coercion (e.g. ISO-8601 → `%DateTime{}`) was lost inside closed maps. Delegates to the existing `{:map, fields}` clause, matching the pattern used in `renderer.ex`, `signature.ex`, and `coercion.ex`.
- 3 regression tests covering the exact shapes from the issue

Closes #932

## Test plan
- [x] `mix precommit` passes (all 4927 tests + 200 demo tests)
- [x] New test: dual-key bypass with string alias is rejected
- [x] New test: dual-key bypass with atom alias is rejected
- [x] New test: `:datetime` inside `{:closed_map, ...}` gets ISO-8601 → `%DateTime{}` coercion
- [x] All existing closed_map and validator tests pass

## Fix Automation State

Fix attempts: 2/3

### DIFF
--- a/lib/ptc_runner/sub_agent/loop/json_handler.ex
+++ b/lib/ptc_runner/sub_agent/loop/json_handler.ex
@@ -123,6 +123,9 @@ defmodule PtcRunner.SubAgent.Loop.JsonHandler do
 
   @doc "Atomize value based on expected type."
   @spec atomize_value(term(), term()) :: term()
+  def atomize_value(map, {:closed_map, fields}) when is_map(map),
+    do: atomize_value(map, {:map, fields})
+
   def atomize_value(map, {:map, fields}) when is_map(map) do
     field_types = Map.new(fields)
 
--- a/lib/ptc_runner/sub_agent/signature/validator.ex
+++ b/lib/ptc_runner/sub_agent/signature/validator.ex
@@ -141,19 +141,8 @@ defmodule PtcRunner.SubAgent.Signature.Validator do
   # ============================================================
 
   defp validate_map_fields(data, fields, path) do
-    Enum.flat_map(fields, fn {field_name, field_type} ->
-      case get_field(data, field_name) do
-        {:ok, _key, value} ->
-          validate_type(value, field_type, path ++ [field_name])
-
-        :missing ->
-          if optional?(field_type) do
-            []
-          else
-            [error_at(path ++ [field_name], "expected field, got nil")]
-          end
-      end
-    end)
+    {errors, _consumed} = validate_map_fields_tracking(data, fields, path)
+    errors
   end
 
   # Like validate_map_fields/3 but also returns the set of consumed data keys
--- a/lib/ptc_runner/sub_agent/signature/validator.ex
+++ b/lib/ptc_runner/sub_agent/signature/validator.ex
@@ -129,7 +129,8 @@ defmodule PtcRunner.SubAgent.Signature.Validator do
   # error, so fields the caller explicitly excluded can't leak through.
   defp validate_type(data, {:closed_map, fields}, path) do
     if is_map(data) do
-      validate_map_fields(data, fields, path) ++ validate_no_extra_fields(data, fields, path)
+      {field_errors, consumed_keys} = validate_map_fields_tracking(data, fields, path)
+      field_errors ++ validate_no_extra_fields(data, consumed_keys, path)
     else
       [error_at(path, "expected map, got #{type_name(data)}")]
     end
@@ -142,11 +143,10 @@ defmodule PtcRunner.SubAgent.Signature.Validator do
   defp validate_map_fields(data, fields, path) do
     Enum.flat_map(fields, fn {field_name, field_type} ->
       case get_field(data, field_name) do
-        {:ok, value} ->
+        {:ok, _key, value} ->
           validate_type(value, field_type, path ++ [field_name])
 
         :missing ->
-          # Check if field is optional
           if optional?(field_type) do
             []
           else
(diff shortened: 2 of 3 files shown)


andreasronge:
## Fix Summary

✅ **Issue Fixed:** Duplicate logic between `validate_map_fields/3` and `validate_map_fields_tracking/3`

`validate_map_fields/3` in `lib/ptc_runner/sub_agent/signature/validator.ex` now delegates to `validate_map_fields_tracking/3` and discards the consumed keys, removing ~15 lines of duplicated field-matching logic. The two functions now share a single implementation.

PR #963, opened 2026-05-16
fix(spec): keyword-tolerant comparison in spec validator

andreasronge:
## Summary

`mix precommit` fails at the `ptc.validate_spec` step — **and fails identically on `main`** (confirmed by running full `mix precommit` in a clean worktree of `origin/main`). 27–31 of 379 spec examples fail, and the count flaps run-to-run.

This is pre-existing drift from the merged keyword-representation commits (`419c447e` atom-leak fix + `676dd5e5` runtime keyword structs), unrelated to any feature work. A source keyword (`:foo`) now surfaces at runtime as **a plain atom**, **a `%PtcRunner.Lisp.Keyword{}` struct**, or **an externalized binary**, depending on the VM atom table at parse time. The spec validator's strict `==` comparison made pass/fail non-deterministic.

## Change

`SpecValidator.validate_example/2` now compares with a new `values_match?/2`:

- **Keyword positions match by name** — atom, `%Keyword{}` struct, and binary are all accepted for a keyword expectation.
- **Type-preserving** — a genuine string expectation still requires a string in the result (the tolerance is keyword-only and guided by the *expected* value's shape, so real string/keyword bugs are not masked).
- Recurses through lists and maps; `%Var{}` matches by name.

The validator checks **spec conformance**, not which equivalent keyword representation the runtime happened to pick.

Also regenerates `priv/spec/checksums.ex` — sections `## 1. Overview` and `## 9. Namespaces…` drifted when the keyword commits edited the spec doc without running `mix ptc.update_spec_checksums`.

## Verification

- `mix ptc.validate_spec` — **379/379, deterministic** (verified over 5 consecutive runs; previously 27–31 failures, flapping)
- `mix precommit` — green (exit 0)
- `mix prepush` (dialyzer) — clean
- New regression tests in `spec_validator_test.exs`: keyword expectations accept any representation; a string expectation still rejects a keyword result.

## Note

The underlying runtime keyword-representation inconsistency (atom vs struct vs binary for the same source keyword, atom-table-dependent) is real but out of scope here — this PR makes the validator robust to it so `precommit` is reliable. Worth a separate look at whether the runtime should settle on one representation.


andreasronge:
Filed #964 to track the underlying runtime keyword-representation inconsistency. This PR makes the spec validator robust to it; #964 is the deeper follow-up to settle the runtime on a single canonical representation.

PR #984, opened 2026-05-17
fix(lisp): make def memory keys deterministic

### LINKED ISSUE #976: [From PR #974] Make def-bound memory keys deterministic
## Context

Raised during review of PR #974 (fix keyword externalization determinism).

## Problem

`externalize_memory_key/1` in `lib/ptc_runner/lisp.ex:758` still uses `existing_atom_or/2` for binary variable names, which consults the VM atom table. This is the same class of non-determinism that PR #974 fixed for keywords.

```elixir
defp externalize_memory_key(name) when is_binary(name) do
  existing_atom_or(name, name)  # still consults VM atom table
end
```

The comment says "for back-compat", and this applies to `def`-bound variable names (not keywords), so it was out of scope for #974.

## Suggested Fix

Make memory keys deterministic by using `SourceAtoms.intern/1` or always keeping them as binaries. Verify that all downstream memory consumers handle both key types before committing to either approach.

## Complexity

Low — mechanical change, but needs verification that downstream memory consumers handle both key types correctly.

andreasronge:
Closes #976.

## Summary
- externalize `def`-bound memory keys through the bounded `SourceAtoms` vocabulary instead of the VM atom table
- keep builtin/source-vocabulary names as atoms while novel user names externalize as binaries
- remove legacy atom aliases when a binary `def` overwrites the same logical binding
- update SubAgent namespace rendering/tests to handle binary memory keys and binary closure names/params

## Notes
- Consulted `docs/guidelines/ptc-lisp-design-guidelines.md` before changing PTC-Lisp behavior.
- Consulted independent high-effort Codex review; kept the duplicate legacy-key fix and rejected broad normalization of untouched initial memory because it broke the existing memory contract.

## Verification
- `mix test test/ptc_runner/lisp/keyword_representation_test.exs test/ptc_runner/sub_agent/namespace/user_test.exs test/ptc_runner/sub_agent/chat_test.exs test/usage_rules_test.exs test/ptc_runner/lisp`
- `mix compile --warnings-as-errors`
- `git diff --check`
- pre-commit hook
- pre-push hook: root tests + dialyzer, mcp_server tests + dialyzer, ptc_viewer tests

## Fix Automation State

Fix attempts: 2/3


andreasronge:
## Fix Summary

✅ **Issue Fixed:** Spec example updated to use binary string keys for user-defined names

In `docs/ptc-lisp-specification.md:4051`, the `memory_after` example previously showed atom keys (`%{high_paid: [...], orders: [...]}`). After the def-key determinism fix in this PR, user-defined names are stored as binary string keys. The example now correctly shows `%{"high_paid" => [...], "orders" => [...]}`.

Also updated `priv/spec/checksums.ex` to reflect the section change.

Issue #987, opened 2026-05-18
Security review: consolidated findings (sandbox limits, data firewall, HTTP transport, supply chain)


andreasronge:
Consolidated findings from a multi-agent security review of `ptc_runner` and `ptc_runner_mcp` (9 topic areas: sandbox escape, resource exhaustion, parser/analyzer robustness, tool-calling boundary, HTTP transport, upstream proxying/SSRF, credentials, trace/debug exposure, supply chain).

**Overall:** architecture is security-conscious — no sandbox *escape* (filesystem/network/parent-state confidentiality breach) was found. Weaknesses cluster around (a) resource-limit enforcement gaps and (b) a misleadingly-named data "firewall."

---

###  ✅  H1 — `pmap`/`pcalls` workers bypass the 10 MB heap cap
`max_heap_size` is a per-process flag and is **not inherited** by `Task.async_stream` child processes. Generated PTC-Lisp using `pmap`/`pcalls` allocates effectively unbounded memory in uncapped children — reproduced **1.1 GB allocated under the default 1 s sandbox timeout** (~100× the documented 10 MB limit). Host-node DoS / OOM. Fixed #1010 
- `lib/ptc_runner/sandbox.ex:94-120`, `lib/ptc_runner/lisp/eval.ex:463-560`
- Fix: set `max_heap_size` on every pmap/pcalls worker (`Process.flag(:max_heap_size, ...)` as first statement, or explicit `spawn` with the option), budget = `max_heap / max_concurrency`.

### ✅ H2 — Parser & analyzer run outside the sandbox with no input-size/depth limit
**Status:** Fixed (#988 / PR #989). Consolidated with H3 + H4 — same root cause — and fixed together.
`Parser.parse` + `Analyze.analyze` run in the caller process *before* `Sandbox.execute`; the timeout/heap limits never cover them. Adversarial program text exhausts host CPU/memory/stack.
- `lib/ptc_runner/lisp.ex:391-453`, `lib/ptc_runner/lisp/fast_parser.ex`
- Fixed: compile phase (parse + analyze + static checks) now runs in a bounded process via `Sandbox.run_bounded/2`; `max_program_bytes` preflight (1 MB); `@max_nesting_depth 64` in the parser; separate `compile_timeout` (5 s). Covers both `run/2` and `validate/2`. Regression test: `test/ptc_runner/lisp/compile_limits_test.exs`.

### ✅ H3 — Unbounded short-fn arity → host OOM
[...]

PR #996, opened 2026-05-18
fix(mcp): harden HTTP DNS rebinding checks

andreasronge:
## Summary
- validate loopback-bound MCP HTTP requests against loopback Host/authority before auth/body handling
- keep missing Origin valid for non-browser clients while rejecting invalid browser Origins
- reject present non-JSON POST Content-Type values and document the behavior

Closes medium issue: https://github.com/andreasronge/ptc_runner/issues/987

## Verification
- mix test test/ptc_runner_mcp/http_router_test.exs test/ptc_runner_mcp/http_config_test.exs
- mix test (from mcp_server/)
- pre-commit: format, compile --warnings-as-errors, Credo, scoped tests
- pre-push: root tests + dialyzer, mcp_server tests + dialyzer, ptc_viewer tests

## Fix Automation State

Fix attempts: 2/3

### DIFF
--- a/docs/mcp-server-configuration.md
+++ b/docs/mcp-server-configuration.md
@@ -91,6 +91,12 @@ HTTP emits sanitized telemetry under `[:ptc_runner_mcp, :http, ...]`
 and logs request stop lines to stderr. Raw bearer tokens and raw
 `MCP-Session-Id` values are not logged.
 
+When the HTTP listener is bound to loopback, `/mcp` also requires a
+loopback `Host`/authority value. This lets non-browser clients omit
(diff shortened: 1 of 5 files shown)


andreasronge:
## Fix Summary

✅ **Issue Fixed:** Added DELETE hostile Host regression test

Added `test "loopback bind rejects hostile Host on DELETE"` to `mcp_server/test/ptc_runner_mcp/http_router_test.exs`. The test sends a `DELETE /mcp` request with `Host: attacker.example` and asserts a 403 response with body `"forbidden"`, providing explicit regression coverage for the DELETE path through `authenticate/2`.

Issue #993, opened 2026-05-18
Security: binary-heavy PTC-Lisp can exceed advertised sandbox memory ceiling


andreasronge:
Follow-up from #987 and related to #988, but this should be fixed in a **separate PR**.

Medium finding: **large binaries allocate off-heap and can exceed the advertised sandbox memory ceiling**.

## Problem

The sandbox currently uses BEAM `:max_heap_size` as the main per-execution memory cap:

- `lib/ptc_runner/sandbox.ex:95` (`Sandbox.execute/3`)
- `lib/ptc_runner/sandbox.ex:228` (`Sandbox.run_bounded/2`)

Large binaries are allocated in BEAM shared-binary storage. Because the current `:max_heap_size` option does not set `include_shared_binaries: true`, binary-heavy programs can allocate far beyond the advertised "~10 MB" sandbox ceiling while the process heap remains small.

Affected producer paths include:

- `str` / `pr-str` / `join` in `lib/ptc_runner/lisp/runtime/string.ex`
- string transformations such as `replace`, `split`, regex split/replace, and grep-style flows
- `json/generate-string` in `lib/ptc_runner/lisp/runtime/json.ex`
- formatting paths that materialize large string output before later preview truncation

Preview/history truncation later in the SubAgent/MCP response path does not fully mitigate this, because the large binary has already been allocated and may be copied or sent back to the parent process.

## Quick reproduction

This currently succeeds despite a tiny heap cap, because the 20 MB binary is off-heap:

```elixir
PtcRunner.Sandbox.run_bounded(
  fn -> String.duplicate("x", 20_000_000) end,
  timeout: 5_000,
  max_heap: 1_000
)
```

Expected security posture: this should either be killed by the sandbox budget or rejected by an explicit binary/result byte limit.

A manual `Process.spawn/2` test with `include_shared_binaries: true` under the same heap cap is killed, so shared-binary accounting materially improves this failure mode.

## Impact

Binary-heavy PTC-Lisp can violate the advertised memory budget and put pressure on VM-wide memory:

- large returned binaries may be allocated outside the child heap budget
- large binary results may be sent back to the parent before preview truncation
- result/history truncation can reduce later storage but cannot undo the original allocation
- `max_heap` is currently documented in a way that can be misread as a whole sandbox memory ceiling

This is a process memory-boundary bug, distinct from the compile-phase isolation issue in #988. It should be handled in a separate PR so the memory contract and regression tests stay focused.

[...]

PR #1010, opened 2026-05-19
fix(security): cap pmap/pcalls parallel workers (H1)

andreasronge:
Fixes security finding **H1** of #987: PTC-Lisp `pmap`/`pcalls` parallel workers were not covered by the sandbox heap cap.

## Problem

`pmap`/`pcalls` previously spawned workers through `Task.async_stream`. `max_heap_size` is a **process-creation** BEAM option, and `Task` creates the worker process internally, so there is no hook to apply the flag before the worker heap is created. The closure, including captured values such as a large `let`-bound collection, could be copied onto an uncapped worker heap before any worker code ran.

That made it possible for generated PTC-Lisp to allocate effectively unbounded memory across parallel workers, bypassing the documented sandbox heap cap.

## What This PR Does

Replaces the Task-based `pmap`/`pcalls` paths with an owned worker lifecycle:

- Adds `PtcRunner.Lisp.Eval.ParallelRunner`.
  It owns worker spawn, bounded scheduling, result collection, shared deadline handling, exit classification, and cleanup. Workers are spawned through `Process.spawn` with `{:max_heap_size, ...}` set at creation.
- Adds `PtcRunner.Lisp.Eval.ParallelBudget`.
  It is an `:atomics`-backed global slot semaphore created once per top-level `Lisp.run/2` and shared through nested `pmap`/`pcalls`.
- MCP callers pass explicit worker limits derived from `program_memory_limit_bytes`, so the user-facing MCP program memory limit is not silently multiplied by the in-process `max_parallel_workers` default.
- Uses a **fixed per-worker heap cap plus a global worker slot cap**, rather than dividing the heap budget by local concurrency:

```text
max live parallel heap ~= max_parallel_workers * worker_max_heap
```

Defaults:

- `worker_max_heap` defaults to the sandbox `max_heap`.
- `max_parallel_workers` defaults to `8`.

Both are operator-configurable. If no worker slot is available, nested parallelism fails closed with `:parallel_capacity_exceeded`; it does not silently fall back to sequential execution.

## Guarantees

- Worker heap cap applies at process birth, before closure/captured data is copied.
- Worker heap caps include shared/refc binary accounting via `include_shared_binaries: true`, matching the current sandbox memory model from #995.
- Nested `pmap`/`pcalls` share the same global worker budget and the same absolute deadline.
- Heap kills are surfaced as `:memory_exceeded`.
- Deadline expiry is surfaced as `:timeout`, distinct from heap failure.
- Global worker budget exhaustion is surfaced as `:parallel_capacity_exceeded`.
- Worker slots are released only after worker termination, not just after result delivery.
- `ParallelBudget.release/1` uses a CAS loop and raises on underflow rather than decrementing then clamping. Underflow is treated as a caller bug because clamping can race with a valid acquire and corrupt a hard security semaphore.
- Cleanup covers normal completion, heap kill, timeout, worker crash, spawn failure, and linked-caller cancellation.

## Rebase / Follow-Up Fixes Included

This branch has been rebased onto current `main` (`502b3bcd`) and the earlier review follow-ups have been incorporated:

- Hand-integrated with main's newer `Lisp.run/2` flow around `max_program_bytes`, bounded compile, `do_run_inner/2`, and `execute_eval/2`.
- Added `include_shared_binaries: true` to parallel worker `max_heap_size` options.
- Added shared/refc binary worker regression coverage for both `pmap` and `pcalls`.
- Added MCP-specific parallel limit derivation so `program_memory_limit_bytes` remains an aggregate program budget for parallel workers in stateless and session execution paths.
- Fixed stale worker `EXIT` cleanup ordering so worker exits cannot be mistaken for linked-caller cancellation after spawn failure.
- Replaced decrement-then-clamp `ParallelBudget.release/1` with CAS decrement and visible underflow failure.
- Added deterministic lower-level `ParallelRunner` tests for spawn failure, stale worker exits, slot leaks, deadlines, cancellation, and lifecycle cleanup.
- Added Lisp-level integration coverage for the public contract.

## Tests

Verified locally on this branch:

- Focused H1 suite: `80 tests, 0 failures`
- Full root suite: `5104 tests, 380 doctests, 3 properties, 0 failures`
- `mix credo --strict`
- `mix dialyzer`
- `mcp_server`: `mix format --check-formatted`
- `mcp_server`: targeted tests for sandbox/session limit plumbing, `44 tests, 0 failures`
- `mcp_server`: `mix credo --strict`
- `mcp_server`: `mix dialyzer`
- Pre-commit hook passed during amend

New or expanded coverage includes:

- `pmap` worker heap cap -> `:memory_exceeded`
- `pcalls` worker heap cap -> `:memory_exceeded`
- captured closure data capped at worker spawn
- shared/refc binary memory accounted in workers
- nested `pmap` capacity exhaustion -> `:parallel_capacity_exceeded`
- nested `pcalls` capacity exhaustion -> `:parallel_capacity_exceeded`
- `max_parallel_workers: 1` still permits a single non-nested pmap
- timeout remains distinct from memory exhaustion
- stale worker `EXIT` does not mask a later spawn failure
- worker slots are released after termination
- budget release underflow is visible and does not mutate the counter
- MCP stateless and session execution derive parallel worker limits from the configured program memory limit
- linked-caller cancellation is not swallowed by scoped `trap_exit`

## Known Caveat

`mix prepush` was attempted after installing `mcp_server/` deps. The root project tests and Dialyzer passed under the hook, but the nested `mcp_server/` suite failed in existing stdio fixture tests because `test/support/mock_server.exs` is launched as plain `elixir` and cannot load `Jason` in that subprocess:

```text
** (UndefinedFunctionError) function Jason.decode/1 is undefined (module Jason is not available)
```

That appears unrelated to this H1 change and is not addressed in this PR.

## Scope Notes

This PR is scoped to H1 from #987. Other sibling findings have already landed separately, including the shared-binary sandbox memory accounting from #993/#995, which this PR now mirrors for parallel workers.

Refs #987 (H1).

## Fix Automation State

Fix attempts: 2/3

### DIFF
--- a/docs/ptc-lisp-specification.md
+++ b/docs/ptc-lisp-specification.md
@@ -3843,9 +3843,27 @@ This means `-1` is always the integer negative one, never a symbol named "-1". S
 | Resource | Default | Notes |
 |----------|---------|-------|
 | Timeout | 1,000 ms | Execution time limit |
-| Max Heap | ~10 MB | Memory limit (1,250,000 words) |
+| Max Heap | ~10 MB | Sandbox-process memory limit (1,250,000 words) |
+| Worker Max Heap | = Max Heap | Fixed per-worker `pmap`/`pcalls` heap cap |
+| Max Parallel Workers | 8 | Global cap on live `pmap`/`pcalls` workers |
 | Max Tool Calls | 10 | Per-program tool invocation limit |
 | Loop/Recur Iterations | 1,000 | Per-loop/recur jump limit; ordinary non-tail recursion is bounded by timeout and heap |
+
+Every `pmap`/`pcalls` worker process — top-level *and* nested — is
+spawned with a **fixed** `max_heap_size` of `Worker Max Heap` words (the
+cap is NOT divided by concurrency). The number of such workers alive at
+once, across the whole program and at every nesting depth, is bounded by
+a shared slot budget of `Max Parallel Workers`. Aggregate live parallel
+heap is therefore bounded by `Max Parallel Workers × Worker Max Heap`.
+
+- A worker that exceeds its fixed heap cap is killed; the program fails
+  with `:memory_exceeded`.
+- A `pmap`/`pcalls` that cannot obtain a worker slot — e.g. deeply
+  nested parallelism that would exceed the global budget — fails with
+  `:parallel_capacity_exceeded` (there is no sequential fallback).
+- The whole parallel operation (including nested `pmap`/`pcalls`) shares
+  one deadline derived from `pmap_timeout`; exceeding it fails with
+  `:timeout`.
 
 *Note: Hosts may configure higher timeouts (e.g., 5,000ms) to accommodate slow tool calls.*
 
--- a/lib/ptc_runner/lisp.ex
+++ b/lib/ptc_runner/lisp.ex
@@ -43,6 +43,15 @@ defmodule PtcRunner.Lisp do
 
   alias PtcRunner.Lisp.Eval.Context, as: EvalContext
   alias PtcRunner.Lisp.Eval.Helpers
+  alias PtcRunner.Lisp.Eval.ParallelBudget
+
+  # Default capacity of the global parallel-worker slot semaphore (see
+  # `PtcRunner.Lisp.Eval.ParallelBudget`). Conservative: with the
+  # default ~10 MB (`1_250_000`-word) `worker_max_heap`, the aggregate
+  # worst-case parallel heap is bounded at roughly `8 × 10 MB = 80 MB`.
+  # The top-level sandbox process is NOT counted as a slot. Overridable
+  # per call via the `:max_parallel_workers` option to `run/2`.
+  @default_max_parallel_workers 8
   alias PtcRunner.Lisp.Format.Var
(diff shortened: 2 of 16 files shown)


andreasronge:
## Fix Summary

✅ **Issue Fixed:** Misplaced `§ 9.3` and `No :signature` comments in `mcp_server/lib/ptc_runner_mcp/sandbox.ex`

The two comments were concatenated with the parallel-worker comment block above `base =`, making them appear to describe parallel worker logic instead of their actual subjects.

**Changes made (`mcp_server/lib/ptc_runner_mcp/sandbox.ex`):**
- Added a blank line after the parallel-worker comment block to visually separate it from the `§ 9.3` comment — the Elixir formatter hoists inline keyword-list comments, so the blank line is the stable way to prevent the two blocks from reading as one
- `§ 9.3` comment now immediately precedes `base =` (and thus `strict_data: true` inside the list), clearly describing why that option is set
- `No :signature` comment moved to after the `base =` assignment closes, just before the Phase 4 comment, so it describes the absence of `:signature` in the `Lisp.run/2` call

PR #1018, opened 2026-05-22
fix(lisp): standardize builtin type errors

andreasronge:
## Summary
- Wrap callable Env builtins with metadata so runtime validation has the public builtin name and argument contract
- Add shared Args validation for selected map and collection builtins with canonical `function: arg n expected type, got type` diagnostics
- Preserve arity precedence, raw runtime callables, formatting, registry classification, and custom swapped-argument hints

## Verification
- `mix credo --strict`
- `mix test test/ptc_runner/lisp`
- Pre-push hook: root full suite + dialyzer, mcp_server full suite + dialyzer, ptc_viewer full suite

## Fix Automation State

Fix attempts: 2/3

### DIFF
--- a/lib/ptc_runner/lisp/env.ex
+++ b/lib/ptc_runner/lisp/env.ex
@@ -12,10 +12,12 @@ defmodule PtcRunner.Lisp.Env do
   - `{:collect, fun}` - Collects all args into a list and passes to unary function
   """
 
+  alias PtcRunner.Lisp.Env.Builtin
   alias PtcRunner.Lisp.Runtime
 
   @type binding ::
-          {:normal, function()}
+          Builtin.t()
+          | {:normal, function()}
           | {:variadic, function(), term()}
           | {:variadic_nonempty, atom(), function()}
           | {:multi_arity, atom(), tuple()}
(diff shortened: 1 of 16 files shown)


andreasronge:
## Fix Summary

✅ **Issue Fixed:** Removed dead code `Builtin.binding/1` and unreachable `custom_error` clause for `sort-by`

- **`lib/ptc_runner/lisp/env/builtin.ex`**: Removed `binding/1` function and its `@spec` (lines 32–34). It was functionally identical to `unwrap/1` and had zero callers outside the module.
- **`lib/ptc_runner/lisp/runtime/args.ex`**: Removed the `defp custom_error("sort-by", 3, :seqable, comp, [key, coll, comp])` clause (lines 153–157). It was unreachable because `custom_call_error/2` runs first with strictly broader guards (covers `LispKeyword` keys and uses `valid_callable?` instead of `is_function`).

All 5103 tests pass.

andreasronge:
@claude fix Unreachable specific_type_error(:update_vals, [%Builtin{} ...] clause in helpers.ex (shadowed by Args.validate! which fires first for all Builtin-wrapped calls)

PR #1032, opened 2026-06-01
[codex] fix arithmetic range conformance quick wins

andreasronge:
## Summary
- fix arithmetic quick wins for unary `/`, zero-arity `-`, unary `+`/`*`/`/` validation, signed-zero preservation, `parse-long` Java long bounds, and `int` NaN/char/overflow coercion
- handle bounded `(take n (range start end 0))` while preserving left-to-right evaluation and builtin shadowing semantics
- turn nil/nonnumeric `range` bounds into type errors and reclassify stale `GAP-S69` floating zero-division backlog item after Clojure verification
- update conformance inventory/audit docs, function metadata, spec checksum, and regression cases

## Review
- high-effort Codex review found evaluation-order, HOF error-class, stale GAP-S69 status, and signed-zero issues; all were addressed before commit

## Validation
- `mix test test/ptc_runner/lisp/runtime_arithmetic_test.exs`
- `mix ptc.conformance_report --write-inventory`
- `mix precommit`
- pre-push hook: root tests + dialyzer, `mcp_server` tests + dialyzer, `ptc_viewer` tests

## Fix Automation State

Fix attempts: 2/3


andreasronge:
## Fix Summary

✅ **Issue Fixed:** Extracted duplicated unary helpers from `apply.ex` and `callable.ex` into a single shared location in `Math`

**What changed:**
- Added `unary_variadic/2`, `unary_plus/1`, and `unary_variadic_nonempty/3` as public functions on `PtcRunner.Lisp.Runtime.Math`
- Updated call sites in `Apply` (lines 185, 202, 235) and `Callable` (lines 86, 98) to delegate to `Math.*`
- Deleted the private copies from both modules (net: 24 insertions, 39 deletions across 3 files)
- Removed the dead `Kernel.-/2` check from the `apply.ex` copy during extraction (it was absent from `callable.ex` and unreachable since `-` routes through `variadic_nonempty`)
- Removed the now-unused `SpecialValues` alias from both `apply.ex` and `callable.ex` (it was only referenced by the deleted private `unary_plus`)

All 5257 tests pass (`mix precommit`).

Issue #1030, opened 2026-06-01
[tracking] PTC-Lisp Clojure conformance: status + open-gap backlog (97 open)


andreasronge:
> Status snapshot. Source of truth: `docs/clojure-conformance-gaps.md` + `docs/conformance-classification-log.md`.

## Summary

**213 active entries** (DIV-17 removed-as-conformant; DIV-47/48/49/50 added) — **open (BUG) 97 · DIV/by-design 57 · fixed 57 · reclassified/matches Clojure 1 · UNSUPPORTED 1**.

_Updated 2026-06-02: GAP-J02 (#1033), GAP-S51 (#1034), GAP-S128 (#1035) fixed since the prior snapshot._
_Updated 2026-06-02 (parallel worktree batch, direct to main): **GAP-S09** (find associative lookup), **GAP-S22** (get-in present-nil), **GAP-J14** (`.substring` finite-float indexes) fixed; **GAP-S136 → DIV-49** (`map-entry?` has no distinct MapEntry type), **GAP-S74 → DIV-50** (`split` signals on multi-char plain-string delimiters) reclassified; **DIV-48** minted (`find` on non-associative collections returns a recoverable `:type_error`). 5 subagents in isolated worktrees, integrated serially; full suite green, codex-reviewed._
Regression coverage: **211/214** documented GAP/DIV ids have a conformance case (**1,178** cases, run vs Babashka). The three documented ids without regression cases are reclassified/no-longer-bug entries: GAP-S10, GAP-S63, GAP-S69.

**Open BUGs by priority:** P0 = 0 (GAP-S63 resolved → DIV-47) · P1 = 11 · P2 = 84 · P3 = 2.

### Reclassified 2026-06-01 (14)

- **char ≡ one-char-string value model — 7 BUG → DIV** (joins DIV-35/40/41): GAP-S44, S116, S120, S125, S126, S130, S133.
- **value-model BUG → DIV (3):** GAP-J20 (temporal unification), GAP-S141 (flat namespace/DIV-07), GAP-S122 (single-precision float range).
- **BUG → UNSUPPORTED (1):** GAP-J18 (`Instant.toEpochMilli` unimplemented).
- **DIV → BUG (2):** DIV-14 (destructuring in `if-let`; cf GAP-S145), DIV-33 (no rule).
- **Removed (1):** DIV-17 (nested `#()` — conformant, both reject). · **Rationale added:** DIV-04/05.
[...]

PR #1033, opened 2026-06-01
fix(lisp): implement Java Boolean parse semantics

andreasronge:
## Summary
- Split `Boolean/parseBoolean` from generic `parse-boolean` so Java interop matches Java semantics for strings: case-insensitive `"true"` returns true and other strings return false.
- Preserve Java null behavior by returning false for PTC `nil`, while non-string, non-nil inputs raise a type error.
- Convert GAP-J02 conformance cases to fixed regressions and update generated docs, audits, and spec checksums.

Fixes GAP-J02.
Refs #1030.

## Verification
- `mix test test/ptc_runner/lisp/double_constants_test.exs`
- `mix test test/ptc_runner/lisp/double_constants_test.exs test/ptc_runner/lisp/registry_test.exs`
- `mix precommit`
- pre-push hook: full root tests, root Dialyzer, `mcp_server/` tests, `mcp_server/` Dialyzer, `ptc_viewer/` tests
- independent high-effort Codex review; addressed stale registry fallback finding

## Fix Automation State

Fix attempts: 2/3


andreasronge:
## Fix Summary

✅ **Issue Fixed:** Added bidirectional `see_also` cross-reference between `parse-boolean` and `Boolean/parseBoolean`

In `priv/functions.exs`, updated the `parse-boolean` entry's `see_also` field from `["parse-long", "parse-double"]` to `["parse-long", "parse-double", "Boolean/parseBoolean"]`. The `Boolean/parseBoolean` entry already listed `see_also: ["parse-boolean"]`, so this makes the cross-reference bidirectional — users and LLMs can now discover the relationship from either direction.

PR #1170, opened 2026-08-03
feat(cli): add stable preparation and provider session checkpoint

andreasronge:
## Summary

This is Checkpoint A of the stable CLI plan. It delivers slices 0 through 5d2b as a mergeable foundation instead of holding all remaining CLI, doctor, publication, REPL, and packaging work in one branch.

- seal transport-neutral run requests, command preparation, diagnostics, limits, and provider identity;
- make installation catalogs inert until explicit runtime activation;
- introduce one bounded provider-session owner with scoped LIFO cleanup and registered process/port roots;
- run active provider validation and optional application admission behind the provider-activity boundary;
- cut `mix ptc.run` over to inert destination preflight followed by one active provider session;
- stop the core OTP application from implicitly starting `req_llm`/`llm_db` or allowing provider-owned dotenv loading; and
- split the remaining plan into sequential Checkpoints B through G.

## Why

The first slice-5 implementation combined too many lifecycle and hosted-service concerns into an unreviewable change. This branch replaces that draft with small owner-backed boundaries and merges the useful safety improvements before the complete stable CLI is finished.

## Impact

Existing `mix ptc.run` commands now reject deterministic destination failures before provider activity, admit provider applications explicitly, and clean provider resources through one session. Provider-free commands do not start optional provider applications. Downstream Mix projects can depend on `req_llm` without the command starting the caller's whole application before admission.

Credential resolution and provider acquisition still use the transitional runtime registry. Bounded credential resolution, process-local OAuth/acquisition composition, doctor, publication, command/REPL parity, standalone packaging, and final acceptance remain in follow-up checkpoints documented in the plan.

## Validation

- `mix precommit`
  - 5,421 root tests passed
  - 71 Viewer tests passed
  - 36 launcher tests passed
  - formatting, compilation, Credo, specification, schema, generated-doc, semantic-projection, and package gates passed

## Review status

No fresh cumulative PR review was run before opening this draft. External review is intentionally pending and must complete before merge.


andreasronge:
## Review notes from a live checkout

Built two worktrees — merge-base `612621d2` vs PR head `c79aa572` — and ran everything below on both. Full root suite on the PR: **5421 passed**, so the number in the PR body reproduces independently.

Verdict: solid, well-tested architecture work that lands a genuinely useful boundary, but it ships two behavior changes that will bite users, and the docs only cover one of them.

### What it enables

**1. The core no longer drags provider libraries into the VM.** `{:req_llm, optional: true, runtime: false}` removes it from the inferred startup set:

```
                                        main      PR #1170
inferred_startup_set_includes_req_llm:   true       false
running_after_core_start[req_llm]:       true       false
running_after_core_start[llm_db]:        true       false
```

On main, anything that starts `:ptc_runner` transitively starts `req_llm` *and* `llm_db`. That's the real prize for embedders.

**2. Preparation is genuinely inert.** With `OPENROUTER_API_KEY` unset entirely, phases 2–5 complete on a provider-bearing manifest:

```
prepare_succeeded_without_credential: true
selected_providers: ["deepseek"]
provider_activity: false
req_llm_started: false   llm_db_started: false
process_delta: 1                          # just the activity marker
filesystem_paths_in_prepared_run: []      # searched the serialized struct
credential_material_in_prepared_run: []
--- phase 8 (open active session) ---
provider_activity: true
req_llm_started: true    llm_db_started: true
```

Phase 8 is the exact line where activity flips and provider apps start. "Path-free" holds — I grepped the serialized `PreparedRun` for the absolute example dir, the manifest dir, and credential needles; only the *logical* component name `extract.clj` survives, which is by design.

**3. Traces stop leaking the raw model selector.** Same live run, `run-started.connector_snapshots`:

```
main:  "model": "openrouter:deepseek/deepseek-v4-flash", "params": {}, "cache": false
PR:    "declaration": {"installation_revision": "tutorial-llm-v1", ...},
       "acquisition_identity_hash": "314eb6b6..."
```

Labels were already hashed on both branches; the connector snapshot was publishing the selector in plaintext right next to them. Real fix. The viewer JS handles both shapes via `connector.acquisition || connector`.

**4. Live runs still work.** Real deepseek call through the new prepare → `ProviderActiveSession` → `RunBuilder` path, with `--trace` and `--output`, returned the correct extraction and a clean 6-event trace.

### Problems

**`.env` credentials silently stop working — and three docs still recommend them.** Same command, same `.env`, key not exported:

```
main:  {"value":{"model_output":"{\"project\":\"Atlas\",...}"}}   exit 0
PR:    ** (Mix) ptc.run failed: :credential_unavailable
```

`req_llm` was the thing loading `.env`; it now starts with `load_dotenv: false`. `docs/guides/host-configuration.md` was correctly rewritten for this. But `docs/guides/building-agents.md:257`, `examples/kernel-tutorial/README.md:29`, and `examples/viewer-demo/README.md:8` still tell users to put `OPENROUTER_API_KEY` in `.env` — following the tutorial now yields `:credential_unavailable`.

**Cold start roughly doubled, despite starting fewer applications.**

```
cold core start:   main 156–174 ms      PR 318–356 ms
```

`SemanticRevision.current()` was added to `Application.start/2` and costs **319 ms on first call** (20 ms without it; 0.003 ms cached). Every `mix ptc.run`, every REPL, and every embedder pays it at boot — and it is fixed latency the planned standalone `ptc` binary will inherit.

**`:command_vm` makes `mix ptc.run` single-shot per VM.** The Mix task hardcodes `:command_vm`, which refuses a provider application it did not start:

```
command_vm, provider app not yet running  -> ok
command_vm, provider app already running  -> active_preflight/provider_application_unavailable
host_owned, provider app already running  -> ok
```

Two `Mix.Tasks.Ptc.Run.run/1` calls in one VM: **main ok/ok, PR ok/fail**. The plan documents this constraint, but the Mix frontend exposes no way to opt into `:host_owned`, so `iex -S mix` workflows and any task chaining two runs break with a diagnostic that reads like a bug.

**Scope vs. the plan's own budget.** The plan sets "fewer than 1,500 changed lines in one review" and "fewer than 500 lines in an orchestration module." Checkpoint A is **29,415 additions / 222 files**; `command_contract.ex` (1,134), `provider_session.ex` (854), and `application_package.ex` (654) all exceed the module budget. `CommandEngine` (537 lines, plus a 3,948-line test file) has **no caller anywhere in `lib/`** — merged a full checkpoint ahead of its consumer. The 15 commits are individually reviewable; the PR as a unit is not, which is the same failure mode the PR body says it was created to avoid.

### Minor

The new `priv/schemas/ptc-command-envelope-v1.schema.json` is a **451 KB single-line** file with 414 `oneOf` branches and 19 levels of nesting. It is generated and `--check`-gated so drift is caught, but it is an awkward thing to hand a consumer as "the stable V1 contract."

### Suggested follow-up

The credential-source change, the admission-mode matrix, and the startup cost are all worth locking in as regression tests — the first two are exactly the gates the plan says slice 5 should have.

andreasronge:
## Follow-up: where the doubled cold start actually goes

Digging into the startup regression from my earlier comment. Short version: it is not the hashing, moving the call does not fix it, and about half of it is recoverable cheaply.

### Root cause

`PtcRunner.Application.start/2` calls `SemanticRevision.current()`, which walks the transitive application closure from the 8 audited roots and SHA-256s every regular file under each app's `ebin/` and `priv/`:

```
apps in closure:     54 (45 present on disk)
files hashed:      1940
bytes hashed:      51.7 MiB
```

Phase breakdown of `semantic_revision.ex:186-206`:

```
enumerate (Path.wildcard + File.regular?)    99.4 ms
stat only                                    66.5 ms
read only                                    94.2 ms
sha256 of all 51.7 MiB in one pass           64.3 ms
```

All 51.7 MiB hashes in 64 ms. The remaining ~260 ms is per-file syscall overhead across 1940 files. The floor — `open`+`close` on those files with no read and no hash — is **105.7 ms**.

### Three fixes that don't work

**Larger async thread pool.** `thread_pool_size` defaults to 1, so this looked like the obvious culprit. `+A 64`: 335 ms → 353 ms. No effect.

**Parallelism.** Over apps: 257 ms. Over files at concurrency 32: 242 ms, versus 249 ms serial. APFS metadata operations do not scale here.

**Moving the call off the boot path.** This is the one that matters. I patched `application.ex` three ways and measured `mix ptc.run` on a provider-free manifest, 5 runs each:

```
baseline (main)                              min 1121 ms
A. PR as merged (eager, blocking)            min 1319 ms
B. no eager call (lazy on first use)         min 1322 ms
C. background prefetch Task at boot          min 1339 ms
```

Removing the eager call changes nothing for `mix ptc.run`, because `ApplicationPackage` calls `SemanticRevision.current()` on every package acquisition (`application_package.ex:403`, re-verified at `:197`). Prefetching is slightly worse — the task contends with the work it was supposed to overlap. The ~200 ms is structural to running a manifest, not to where the call sits.

### What does work

```
current (2 stats/file, full closure)          249.5 ms
single stat/file, full closure                242.9 ms
current, skip OTP-shipped apps                190.0 ms
single stat/file + skip OTP-shipped apps      139.9 ms   <- -44%
floor: open+close 1940 files                  105.7 ms
```

**Stat each file once.** The traversal currently stats every file twice: `Path.wildcard` → `File.regular?` (stat 1) → `File.stat!` inside `hash_dependency_files` (stat 2) → `open`. A single `:prim_file.read_file_info` yields type, size, and mode together. Pure refactor, no change to what is attested.

**Drop OTP-shipped apps from the closure.** ~25% of the cost, 488 of the 1940 files, and arguably more correct: `runtime_projection/1` already records `erts`, `otp_release`, and `system_architecture`, so byte-hashing `stdlib`, `ssl`, `inets`, `compiler`, and `kernel` re-derives the OTP release identity the expensive way.

Together: **250 ms → 140 ms**. Removing the boot call on top still helps every path that starts `:ptc_runner` without acquiring a package — REPL, tests, provider-free embeddings — where it is 340 ms → 20 ms, and where the computed value is currently discarded.

### Beyond performance

The closure follows `applications`, `included_applications`, and `optional_applications`, which pulls in:

```
req_llm.app applications:           [:zoi, :igniter]
req_llm.app included_applications:  [:llm_db]
llm_db.app  applications:           [:zoi, :igniter]
```

`igniter` is a code-generation dev tool, and it is part of the runtime execution-semantics identity — bumping a codegen dependency changes the revision of every previously built package. `zoi` is 319 files / 92 ms, the most expensive single app in the closure. That reads like a modelling question rather than a cost one.

### Suggested changes

1. Delete the `Application.start/2` call; it discards its result and buys nothing measurable.
2. Stat each file once.
3. Exclude OTP-shipped apps; the release identity is already recorded separately.
4. Reconsider whether compile-time-only applications belong in an *execution*-semantics revision.
5. For the planned standalone binary: a release is a sealed artifact, so the dependency projection can be computed at build time and verified cheaply at runtime instead of rebuilt on every invocation.

Caveat on absolute numbers: measured on macOS/APFS, Apple Silicon, 10 schedulers. ~54 µs per file open is high; Linux is typically an order of magnitude cheaper, so the real cost on CI or a Linux deployment is probably much smaller. Worth confirming there before investing in 2 and 3 — 1 is free either way.

andreasronge:
## Review — Checkpoint A (`c23d919e`)

Reviewed at `c23d919e` against merge-base `0a8f15e9` (157 files, +29,424/−5,682). Findings below were each verified against the code and its callers; the coverage gaps at the end are stated explicitly rather than implied.

**No P1.** Nothing found that corrupts state, leaks a secret outside the VM, or loses a resource permanently. Four items look worth fixing before merge, two of which are contract defects rather than runtime bugs — which matters more than usual for a PR whose purpose is a *stable* contract.

### Framing: the contract has no production frontend yet

`mix ptc.run` references neither `CommandEngine` nor `CommandOutcome`; it aliases only `CommandDiagnostic` and builds its own path via `RunCoordinator`/`RunBuilder`. `CommandOutcome` is referenced nowhere in `lib/` outside `command_engine.ex` and itself.

So the envelope, catalogued diagnostics and `exit_status` machinery are built and heavily tested but not yet load-bearing. That reads as intended scope (the description defers command/REPL parity to a later checkpoint), and it lowers the risk of the new-module count considerably — contract-shaped defects are cheap to fix now. Worth making explicit in the description, since the diff size otherwise implies a larger blast radius than exists.

One consequence today: `ptc.run` ships two error contracts. Option-level failures return bare atoms (`:conflicting_result_destinations`, `:conflicting_mission_inputs`, `:check_has_output_options`, `:authorization_requires_host_config`) that bypass `DiagnosticCatalog` and reach the generic `raise_failure` clause at `lib/mix/tasks/ptc.run.ex:527` to be `inspect`ed into the message. No leak — `CommandDiagnostic` is a closed struct — but `Mix.raise` always exits 1, so a diagnostic's `exit_status` is never applied.

### Worth fixing before merge

**1. `HostRuntimePayload` looks like an encryption boundary but provides no confidentiality.** `lib/ptc_runner/kernel/host_runtime_payload.ex:139`

The AES-256-GCM key is derived by HMAC'ing a constant atom under the module's `:persistent_term` attestation secret, so any code in the same VM can reconstruct it; nonce, ciphertext, tag and `binding` are all public struct fields, and `@opaque` is a Dialyzer annotation with no runtime effect. The sealed plaintext genuinely contains secrets — the schema permits `credentials.*.literal` up to 65536 chars, read at `host_config.ex:583`.

Unlike `Attestation`, which honestly disclaims in-VM confidentiality in its `@moduledoc`, this module is `@moduledoc false` and states no threat model. The sharpest consequence is operational: `:persistent_term` is never garbage collected and is written verbatim into `erl_crash.dump`, so a single crash dump would contain both the key material and every sealed payload.

Suggested: either give it a `@moduledoc` stating the same threat model `Attestation` does, or drop the AEAD layer and rely on struct opacity plus the existing MAC.

**2. `close/1` is documented idempotent but silently no-ops cross-process.** `lib/ptc_runner/kernel/prepared_run.ex:159`, `lib/ptc_runner/kernel/command_preparation.ex:99`

Both are `@doc`'d as idempotent releases and typed `:: :ok`. They delegate to `ProviderActivity.stop/1`, which discards the server's reply — and the server only honours `:stop` from the current controller:

```elixir
defp handle_operation(:stop, {caller, _tag}, %{controller: caller} = state),
  do: {:stop, :normal, :ok, state}

defp handle_operation(:stop, _from, state), do: {:reply, :ok, state}   # replies :ok, stays alive
```

Since `consume/1` deliberately transfers controllership, a creator that hands a prepared run to another process cannot release it and cannot detect the failure. Mitigating: `transfer/3` monitors the new controller, so the marker dies with the consumer rather than leaking permanently.

This is untested in the matching shape. The only `Process.alive?(owner)` assertions are `test/ptc_runner/kernel/command_engine_test.exs:2646` and `:2649`, both same-process. The cross-process test at `:2780` ends with a bare `assert :ok = PreparedRun.close(prepared)` at `:2804`, which cannot fail — `close/1` returns `:ok` unconditionally, and by then the winning `Task` has exited and its monitor already stopped the marker.

Suggested: either let the original creator stop a transferred marker, or return `{:error, :not_owner}`; and correct both `@doc`/`@spec` pairs.

**3. `input_contract_failed` is overloaded across three distinct causes.** `lib/ptc_runner/kernel/execution_input.ex:33` → `lib/ptc_runner/kernel/command_engine.ex:422`

`ExecutionInput.new/3` funnels non-object shape, `StrictJSON.admit` failure and unsealed-contract into one bare `{:error, :invalid_input}`, which projects to the published code `input_contract_failed` with a `nil` path. An application declaring *no* input contract can therefore emit it — e.g. a manifest with no `contracts` section and an input file containing `[]` — sending an operator hunting for a schema that does not exist. The behaviour is currently locked in by `command_engine_test.exs:3462`.

Suggested: return a distinct reason for the non-admissible case with its own projection, reserving `input_contract_failed` for `validate_contract/2` rejections, which are the only ones carrying a classification.

**4. Double-close race can replace a real error with `:provider_cleanup_failed`.** `lib/ptc_runner/kernel/run_builder.ex:925`

`build_active/4`'s error path runs `close_live_session(session)` on a session that `finish_provider_session/6` (`:646`) has usually already closed, guarded only by `Process.alive?/1`. `gen_server` sends the `{:stop, :normal, reply, state}` reply after `terminate/2` but before exit, so in that window a second `ProviderSession.close/1` returns `{:error, :provider_cleanup_failed}` and `prefer_cleanup_error/2` promotes it over the genuine reason (e.g. `:provider_dependency_unavailable`). The window is tiny and there is no deterministic reproduction, but the result is a nondeterministic published error code.

Note the outer close is **not** redundant — errors raised in `build_active`'s own `with` never reach the inner close — so the fix is an explicit closed-flag rather than deletion.

### Hardening

- **`lib/ptc_runner/kernel/provider_activity.ex:200` has no catch-all `handle_info/2`**, and is the only one of these owners using `GenServer.start_link`, so it is linked to its creator and does not trap exits. An unmatched message would raise `FunctionClauseError` and propagate the exit to the process running the command. `use GenServer`'s default `handle_info/2` is `defoverridable`, so the single `:DOWN` clause replaces the log-and-continue default rather than adding to it. No reachable trigger was found — `transfer/3` uses `demonitor(_, [:flush])`, exits are signals not messages without `trap_exit`, and OTP 24+ reply aliases discard late replies — so this is a defensive gap, not a live bug. Sibling owners `provider_session.ex:556` and `run_state.ex:694` both have the catch-all; `host_installation_owner.ex:325` has the same gap with a smaller blast radius. One line each.
- **`run_builder.ex:226` and `:203`** pass the caller's raw `opts` to `execute_built/2` while preflight ran on the anchored copy `assemble/5` computes and discards, so the no-clobber/identity check can be performed on a different file than the one written. Contradicts the module doc's "anchored once before preflight and the same absolute paths are retained". Masked today because `mix ptc.run` pre-anchors.
- **`provider_session.ex:202` `attach_provider/2`** can return `{:error, :provider_session_unavailable}` (catch-all at `:495`, whenever `lifecycle_bound?` is still false), outside both its own `@spec` and `RunState.attach_provider/2`'s `case` → `CaseClauseError` in the dispatching process rather than a fail-closed release. Unreachable today since both live paths bind first. It is also the only public `ProviderSession` function without a non-struct fallback clause.
- **`attestation.ex:40`** guards node-local `:persistent_term` with a cluster-wide `:global.trans/2` (infinite retries) on the run hot path. The lock is genuinely needed — two racing generators would each `put`, invalidating handles attested under the loser — but it only needs to be node-local. A netsplit could stall sealing. Generating owner secrets once at application start would remove both the lock and the lazy `put` from the run path.
- **`application_package.ex:284`** propagates declared-input read/decode failures bare, so a malformed input file is attributed to the manifest (`source: {kind: "application", name: "ptc.json"}`) rather than to the input document — unlike every other referenced document.
- **`manifest.ex:1187`** bounds `$schema`, `events.run_id` and `events.trace_id` with `maxLength` (code points) in the generated schema but `byte_size/1` in the loader, so a schema-valid manifest with non-ASCII values in those three fields is rejected at load. They are the only bounded strings without an ASCII-restricting pattern.
- **`docs/guides/host-configuration.md:562` and `limits.ex:33`** both state the three installed-only timeouts are catalogued ahead of the code that consumes them and are "not yet imposed". All three are consumed in this PR: `doctor_connectivity_timeout_ms` at `host_installation.ex:851`, `provider_cleanup_timeout_ms` at `provider_session.ex:76,111,298`, `selection_validation_timeout_ms` at `provider_active_session.ex:129`. The guide also contradicts itself at line 169.
- `attestation.ex:23` `secure_compare/2` is data-independent and correct, but `:crypto.hash_equals/2` is the idiomatic primitive.

### Verified clean

- **`c23d919e` is a genuine fix.** Both types stay `@opaque`, validation moved into the owning modules, and `ProviderSession.alive?/1` is strictly stronger than the inline `%ProviderSession{pid: pid}` + `Process.alive?` it replaced. Each function head that lost its struct pattern is gated downstream: `build_active/4` via `compatible_limits?/2`'s `false` fallback, `start_provider_tracker/2` via `valid_provider_session?` at `run_state.ex:88`, `use_provider_session/2` via its retained inner `valid?`.
- **`Attestation` delivers what it claims.** HMAC-SHA256 over `term_to_binary(_, [:deterministic])` with a per-owner random key; payload coverage is complete in both callers — `provider_session.ex:842` attests all five mutable fields and `valid?` pins the key set, so a tampered `cleanup_timeout_ms` cannot slip past `compatible_limits?`. The `@moduledoc` correctly scopes it as construction/mutation detection, not an in-VM security boundary.
- **No limit lost, no default silently changed** across the `LimitCatalog` refactor; all 23 pre-existing fields' compiled and installed defaults match the base, and every catalog row has a live consumer. The one behaviour change is a new upper bound, a tightening.
- **No manifest validation dropped** in the 210-line deletion — the rewrite adds checks (component id/library regex, logical-name grammar, dependency shape/sortedness, closed `limits` key set, explicit-`null` rejection). Loader is nowhere more permissive than the schema.
- **`StrictJSON` is genuinely strict** — duplicate keys rejected pre-conversion, depth/node/timeout/heap bounds applied to trusted admission as well as decoding, and zero atom creation from untrusted input.
- **Content digests are canonical** — domain-separated, record-count-prefixed, every field length-prefixed, sorted with duplicate rejection before sorting.
- **Provider cleanup is leak-free** across creator crash between claim/consume, prepare/execute, and force-kill; slot accounting agrees across `cleanup_slots/2` and both `close_roots` arities.
- **No `Agent.get`-then-`Agent.update`** anywhere in the reviewed code; all owner state mutates inside a single callback.
- **OAuth secret containment** — every secret-bearing struct has a `defimpl Inspect` returning `<redacted>`, the GenServers holding secrets override `format_status/1,2` (so crash reports and `:sys.get_status` are covered), and there are zero `Logger.` calls under `mcp_oauth/`.
- **Test conventions clean** — zero `Process.sleep` and zero `:timer.sleep` across all 57 touched test files. `PreparedRun` single-use is properly enforced and tested, including a concurrent two-task test asserting exactly one winner.

### Not covered by this review

Stated explicitly so the gaps are not mistaken for clean results:

- Command envelope conformance against `priv/schemas/ptc-command-envelope-v1.schema.json`.
- `DiagnosticCatalog` closure in both directions (engine-producible codes absent from the catalog; catalog entries nothing can produce).
- `TypedCanonicalJSON`/`SemanticRevision` canonicalisation determinism — relevant because an unstable canonical form would make `priv/semantic_build_projection.json` flap.
- OAuth concurrency: token refresh races, expiry clock semantics, cleanup racing an in-flight request, cross-request context bleed. Only the handoff-dedup question was examined, and `drop_worker/2` dropping `handoffs` alongside a killed worker appears to be deliberate fail-safe behaviour rather than a defect.
- Systematic guide-vs-code drift, particularly the +522 lines in `docs/guides/kernel-maintainer.md`.
- Private-value containment was checked in `ptc.run`'s own report path (which requires `--private-output` and emits only class plus a projected usage) but not through trace/inspection destinations.

andreasronge:
## Review, part 2 — the gaps from the previous comment (`964a0e49`)

Follow-up to the [earlier review](https://github.com/andreasronge/ptc_runner/pull/1170#issuecomment-5168202408), which listed four areas as explicitly not covered. Those are now reviewed at `964a0e49`, which also picked up three commits that arrived after the first pass (`d6307a25`, `770e18e7`, `964a0e49`).

Headline: **envelope conformance, catalog closure and digest determinism all came back clean** — structurally so, in ways worth recording. The new findings are mostly around them: a missing CI gate, and the CLI discarding diagnostics the contract carefully builds.

### Clean results worth recording

- **Envelope conformance is structurally enforced, not merely observed.** All three constructors funnel through `CommandOutcome.seal/3` (`command_outcome.ex:193`), which re-validates the built map against the same `CommandContract.schema()` that `mix ptc.gen_docs` serialises to the checked-in JSON, failing closed on build error, validate error and exception alike. The emitted key sets match the schema's required lists exactly, and the 19 top-level `oneOf` branches are pairwise disjoint, so `oneOf` cannot mis-fire. `exit_status` is correctly absent as a JSON field and re-derived from the same catalog row on the round-trip.
- **The catalog is closed engine→catalog: 30 of 30 producible `(phase, code)` pairs present, 0 missing.** The closure holds one level deeper — the typed provenance each call site attaches also satisfies the catalog's policy tables, and the per-mode gate admits exactly the pairs reachable in each mode. The 40 rows with no producer are all in phases this checkpoint hasn't landed (destination, execution, publication…), i.e. forward declarations rather than dead error paths. Correction to an assumption in my first comment: a missing pair would **not** degrade silently — `new!/3` raises `MatchError`, rescued into `{:internal, :internal_error}`, so it fails loudly at exit 70 instead of the precise status.
- **The semantic-revision digest is deterministic**, verified empirically rather than by reading: the source-closure and path-dependency framing was reimplemented independently and reproduced both checked-in hashes byte-for-byte, with the projection current at `964a0e49` (274/274 sources, zero drift). Map keys are byte-sorted at every level, integers encode as exact decimal and floats as IEEE-754 hex bits (so `1` and `1.0` differ, bignums are exact, `-0.0 ≠ 0.0`), invalid UTF-8 cannot enter, and the single `term_to_binary` carries `[:deterministic]`. Nothing ambient reaches the digest — notably `Context.default_pmap_max_concurrency()` looks ambient but is a compile-time attribute, and a test explicitly flips `:erlang.system_flag(:schedulers_online, ...)` and asserts the revision holds. `770e18e7` changed only *when* the value is computed, not what feeds it.
- **The changed guides are unusually accurate.** ~50 falsifiable claims sampled across all six files — names and arities, struct fields, numeric defaults and ranges, regex grammars, domain-separation strings, error codes, exit statuses, sequencing — all held except the two below. No file in `lib/` links to `docs/plans/`.

### Findings

**1. CI never runs the semantic-revision staleness gate.** `.github/workflows/test.yml` / `mix.exs:126` — P2

Every gate in the `precommit` alias is mirrored as an individual CI step except `mix ptc.gen_semantic_revision --check`; `grep -rn gen_semantic_revision .github/ scripts/ .githooks/` is empty. Its sibling `ptc.gen_docs --check` sits on the adjacent line at `test.yml:287`.

Because `@build_projection` only recompiles when the projection file itself changes, a PR can edit anything under `lib/ptc_runner/` without regenerating, and `SemanticRevision.current()` keeps returning the pre-change digest with CI fully green — the revision attesting a source closure that no longer exists. This PR shows the manual rhythm the gate exists to remove: the projection needed hand-refreshing twice (`1af53948`, and again inside `770e18e7`). One line next to the existing `gen_docs --check` fixes it, and matches the repo's own stated rule that PR CI runs the same checks as individual steps.

**2. The CLI discards catalogued diagnostics wholesale.** `lib/mix/tasks/ptc.run.ex:557` — P2

`raise_failure/1`'s generic clause `inspect`s any `%CommandDiagnostic{}` that isn't `{:provider_declaration, :provider_unknown}`, so the catalog's message and `exit_status` are both thrown away.

Concretely, for a workflow component that fails to compile, `RunCoordinator.compile_required/1` builds a `{:bundle, :compile_failed}` diagnostic with a `:component` source, and the user sees a raw struct dump — additionally truncated mid-struct, because the struct has 11 fields and `inspect` is called with `limit: 10`. The catalogued message never appears, and the shell gets exit 1 rather than 3, so scripts cannot distinguish an application-phase failure from a runtime-limit failure. Same for the four `active_preflight` codes arriving via `ProviderActiveSession.open/3`.

This is distinct from the bare-atom bypass in my first comment: here the value *is* a fully catalogued diagnostic and the catalog is still bypassed, so fixing the atoms would not fix this. A `%CommandDiagnostic{}` clause rendering `diagnostic.message` and exiting `diagnostic.exit_status` handles both, and makes the `provider_unknown` special case redundant.

**3. The revision generator enumerates the working tree, not tracked files.** `lib/mix/tasks/ptc.gen_semantic_revision.ex:135` — P2

`semantic_files/2` uses `Path.wildcard(..., match_dot: true)` filtered only by a hand-maintained exclusion list, so it is not gitignore-aware. A stray `.DS_Store`, vim swapfile or scratch `.ex` under `lib/ptc_runner/` makes `--check` report stale with nothing actually changed — and regenerating bakes the machine-local path and hash into the checked-in expected value. With finding 1 in place, nobody notices until another developer runs `mix precommit`. The launcher root spec having to name `.formatter.exs` and `.gitignore` explicitly is the same mechanism already leaking. Intersecting with `git ls-files`, or failing loudly on an untracked hit, defines the closure by version control rather than by a list that must grow for every new local-artifact name.

**4. `d6307a25` walks back a stated isolation property, and the commit message overclaims.** `provider_application_gate.ex:67`, `ptc.run.ex:318` — P2

The two modes differ in a way the new mode-selection makes load-bearing. `:command_vm` (`:78`) errors if the target is already running, then sets `load_dotenv: false` for `:req_llm` and `:llm_db` before starting them. `:host_owned` (`:67`) requires it to already be running and never touches `load_dotenv`. The new `provider_application_mode/0` picks `:host_owned` whenever `:req_llm` is already running.

So the message's "while keeping dependency dotenv readers disabled" holds only on the `:command_vm` path — in `:host_owned`, exactly the mode this commit adds for repeated invocations, PtcRunner never applies it. (Arguably unavoidable once the app has started, but then the claim shouldn't be unconditional.) Relatedly, `:command_vm`'s documented "rejects an inherited target" guard is now unreachable from `mix ptc.run`, since the mode is chosen precisely when a target is inherited: a fail-closed check became auto-satisfied. Defensible where the surrounding project is trusted, but the gate's `@moduledoc` still presents the rejection as operative.

**5. The `.env` documentation understates its blast radius.** `docs/guides/host-configuration.md:112` — P2

Nothing in the paragraph is literally false, but the gate is per-credential while the effect is not. `Dotenv.load/0` starts at `File.cwd!()` and recurses to `/`, then sets **every** `KEY=VALUE` line via `System.put_env`, latched in `:persistent_term` so the injection is VM-wide and permanent. Two paragraphs earlier the same section promises "There is no ambient provider-specific environment lookup" — a reader will conclude the read is scoped to the one declared credential. It matters more now that `running-and-debugging.md:32` recommends `iex -S mix` and task chaining: the latch means the first invocation's `.env` keeps applying to every later one in that VM. One sentence naming the scope, the upward search and the lifetime would close it. (`d6307a25` did correctly delete the now-false "not an implicit credential source" clause.)

Same paragraph, separately: `host-configuration.md:170`'s `doctor --connect` text is present-tense and unlabeled, but no shipped frontend exposes a doctor command — the only `{:doctor, :connect}` implementation has no caller in `lib/`. One rewrite in future tense fixes both this and the timeout-enforcement contradiction from my first comment.

### Smaller items

- **`run_builder.ex:944`** — `execute_built/2` applies `disclosable/3` only on the persistence-failure branch, so the success branch returns `%Result{value: <live private value>}`. Containment currently rests entirely on `report/3` never touching `.value`; any other caller of the `*_with_class` API that matches `{:ok, result, _class}` and prints or persists `result.value` publishes it with no guard tripping. Redacting on the success path too would make `:redacted` the only private value that leaves the function. (P3 today, because the CLI is the only caller.)
- **`command_engine.ex:460`** — the `application_projection/2` catch-all maps unrecognised reasons to `:schema_violation`, whose only admitted source kind is `:application`; under the `:component` or `:external_input` roles that pair is catalog-forbidden and `new!/3` raises. Not reachable today, but the guard is invisible: adding one error atom to `ApplicationSource.read/3` silently converts a precise exit-3 diagnostic into exit-70. The contract and `:component_override` roles already have such catch-alls; these two don't.
- **`command_engine.ex:33`** — `prepare/1` can return `{:error, :entropy_unavailable}`, a bare atom with no diagnostic, message or exit status, because no run ref exists to build an envelope around. A caller matching `{:error, %CommandOutcome{}}` hits a `CaseClauseError`.
- **`ptc.gen_semantic_revision.ex:184`** — `term_to_binary(entry, [:deterministic])` doesn't pin `minor_version`, and the default flipped from 1 to 2 at OTP 26. Lock entries are atom-dense, so the same unchanged `mix.lock` yields a different `content_identity` per OTP major: `--check` reports stale on an older OTP and the file flip-flops between contributors. Contained only because `mise.toml` and CI both pin 29.0.3.
- **`semantic_revision.ex:210`** (from `964a0e49`) — size and mode now come from a stat taken during whole-closure enumeration while bytes are read later, and the read length is never checked against the framed size. The window widened from microseconds to the full scan (~1,940 files / 51.7 MiB). Not reachable in an ordinary build, but a concurrent `mix deps.compile` would mix a stale size with fresh content. Checking the streamed length against the framed size would close it.
- **`semantic_revision.ex:67`** — `{:ok, encoded} = TypedCanonicalJSON.encode(projection)` hard-matches a result that can legitimately be `{:error, :invalid_json}`, since `StrictJSON.admit` runs under a bounded worker that maps `:timeout` and `:heap_exceeded` onto it. `770e18e7` moved the first computation off the quiet boot path onto the first `ApplicationPackage` validation, so under scheduler starvation this now surfaces as a `MatchError` inside a request.
- **`command_contract.ex:174`** — `$defs/diagnostic` and `$defs/artifact_state` are emitted into the published schema but `$ref`-ed by nothing, and `$defs/diagnostic` is generated from the full catalog, so it is strictly wider than any branch admits. It is 107 KB of the 451 KB file; a consumer codegenning types from `$defs` derives a Diagnostic type admitting pairs no envelope will carry.
- A house-style question rather than a finding: `Attestation`, `Dotenv` and `SemanticRevision.current/0` all guard node-local `:persistent_term` with a cluster-wide `:global.trans`. Individually minor; all three now sit on the run path, so it's worth one decision rather than three.

### Still not covered

OAuth concurrency — refresh races, failed-refresh state, expiry clock semantics, cleanup racing an in-flight request, and cross-request context bleed. Secret containment in that slice **is** verified clean (every secret-bearing struct has a redacting `Inspect` impl, the GenServers holding secrets override `format_status/1,2`, and there are zero `Logger.` calls under `mcp_oauth/`), and the one handoff-dedup question was examined and came back as deliberate fail-safe behaviour rather than a defect.

andreasronge:
## Review, part 3 — OAuth concurrency (`964a0e49`)

Closes the last gap listed in [part 1](https://github.com/andreasronge/ptc_runner/pull/1170#issuecomment-5168202408) and [part 2](https://github.com/andreasronge/ptc_runner/pull/1170#issuecomment-5168373505). Every area flagged as uncovered has now been reviewed.

The lifecycle is **substantially correct under concurrency**, and the core mechanisms are better than most code in this area. One P2 and two P3s.

### What holds up

- **Adoption is genuinely serialized.** Every mutation of `workers`/`monitor_refs`/`handoffs`/`pending_handoffs`/`handoff_pids` happens inside a single `handle_call`/`handle_info` clause — no read-modify-write across two calls. Concurrent adopters of the same `{manager_pid, registrar}` join one in-flight handoff's waiter list and all receive the same normalized result, so no duplicate `handoff_root` is issued. That matters because `ProviderScopeOwner.drop_root/2` is deliberately **not** idempotent and answers a repeat with `{:error, :provider_scope_unavailable}`.
- **Refresh coordination is correct.** `reserve_refresh` elects one leader per generation and parks concurrent callers as waiters rather than issuing duplicate token requests; waiters are answered `:retry` and re-read the store; leader death is covered by `leader_ref`'s monitor; and the store mutation lease's `starting_generation` fence rejects a stale winner. A failed refresh calls `fail_mutation` with explicit `:not_dispatched`/`:possibly_dispatched` provenance and leaves **no usable-looking dead grant** in the store.
- **Clock semantics are clean, and correct for a non-obvious reason.** The store owns the clock: `expires_at_ms = anchor_ms + ttl_ms` is absolute monotonic inside one process, tagged with a `{:time_anchor, serial}` so anchors from another epoch are rejected, and projected to callers only as the `remaining_ttl_ms` duration — `loaded_grant/1` deletes the absolute field before handing the grant out, so no caller can compare it against a wall clock. `unknown_expiry_ttl_ms` is consumed as a TTL, never as an instant. No wall clock anywhere in the path, so there is no skew surface.
- **Cleanup cannot pull a live manager out from under an in-flight request.** The worker *links* the manager rather than monitoring it, so an owner restart propagates death to the credential holder instead of orphaning it; and the reaper answers handoff requests from inside `await_survivors`/`await_termination`/`observe_delayed`, so a manager blocked in its own DOWN handler on `adopt/2` cannot deadlock against a concurrently tearing-down scope.
- **Context bleed is not reachable.** `active` is keyed by caller pid, entries are stamped with a monotonically increasing id, and every late reply re-matches on that id and drops stale results.
- No timers anywhere; monitors are one-per-worker, consumed on DOWN or purged by `drop_worker/2`'s `Map.reject` — which is what stops a queued stale DOWN from evicting a freshly restarted worker.

### Findings

**1. Competing timeouts on the seal path silently suppress token-manager adoption.** `lib/ptc_runner/kernel/mcp_request_context.ex:90` — P2

`safe_call(pid, :seal)` takes `safe_call/3`'s default `5_000` ms, while the handoff that call triggers inside the callback is itself bounded at `ProviderScopeOwner.@request_timeout_ms`, also `5_000` ms. Not merely competing — **exactly equal**, so under a stall both expire at essentially the same instant. `ManagerCleanup`'s moduledoc states the invariant this violates: *"Callers add no competing timeout; each registrar handoff remains bounded by …"*.

When the seal loses that tie:

```elixir
seal_result = safe_call(pid, :seal)

adoption_result =
  if seal_result == :ok,
    do: adopt_authorization(authorization, registrar),
    else: :ok          # adoption skipped — and reported as success
```

`adopt_authorization/2` is skipped entirely and reported as `:ok`. The manager gets no cleanup worker and no handoff; `close_authorization/3` falls through to a direct `TokenManager.close/1`, which returns `{:error, :persistence_failed}` while the manager stays alive. The manager is still a scope root, so the reaper force-kills it — and the unsettled persistence retry this PR exists to preserve is lost. The stall condition is real enough that the suite already simulates it with `:erlang.suspend_process/1` (`token_manager_test.exs:929`). Separately the `{_seal, :ok}` branch turns an otherwise clean close into `{:error, :mcp_transport_error}`.

Fix: call `:seal` with `:infinity` — the handoff is already locally bounded, exactly as `ManagerCleanup.adopt/2` does — and decouple adoption from the seal result, since the context's root handoff and the token manager's adoption are independent operations.

**2. The handoff dedup set's lifetime is keyed to the worker, not the manager.** `lib/ptc_runner/kernel/mcp_oauth/manager_cleanup.ex:114` — P3

`handoffs` records registrars whose force-close set no longer contains this manager — a property of the *manager* — but it is cleared whenever the *worker* record is dropped (`:114` on worker DOWN, `:268` in `drop_worker/2`), and `finish_handoff/3` conversely refuses to record it unless a worker is present (`:235`).

Not reachable today: `drop_worker/2` only fires from the `Process.alive?(worker) == false` branch of `ensure_worker/2`, and `ManagerCleanupWorker` links the manager without trapping exits, so a dead worker implies a dead manager — after which `start_worker/2`'s liveness guard returns `{:complete, ...}` without touching the registrar. But the margin is one line wide: if either `ManagerCleanupWorker` or `TokenManager` ever gains `Process.flag(:trap_exit, true)`, worker death stops implying manager death, the dedup entry is wiped, the next `adopt/2` re-issues `handoff_root/2` for an already-removed root, and non-idempotent `drop_root/2` produces a `{:error, :scope_unavailable}` for a manager that was in fact correctly handed off — surfaced to the user as `:mcp_transport_error`.

Fix: monitor `manager_pid` directly and clear `handoffs[manager_pid]` on *its* DOWN; drop the worker-presence condition at `:235`.

**3. Unbounded, unmonitored manager pid set is killed on close.** `lib/ptc_runner/kernel/mcp_oauth/store/memory.ex:96` — P3

`register_manager/2` accumulates raw pids in a `MapSet` with no monitors, and `handle_call(:close, ...)` later sends `Process.exit(pid, :kill)` to every recorded pid that is alive. A long `mix ptc.run` session starts one manager per OAuth-backed MCP acquisition and every one stays in `state.managers` for the store's lifetime. Since `Process.alive?` is a test a *recycled* pid passes, after enough process churn `close/1` can kill an unrelated process that merely inherited a dead manager's pid.

Fix: `Process.monitor/1` each manager at registration and drop it on `:DOWN`, bounding the set to live managers and making the kill list unambiguous.

### On the earlier handoff question

The tentative refutation in part 2 is confirmed, and can be stated more strongly than "deliberate fail-safe": the scenario also requires an interleaving that cannot be constructed. Reaching `drop_worker/2` needs a dead worker whose manager is still alive, but the worker links the manager and does not trap exits, so worker death implies manager death in every path — and the next `adopt/2` then short-circuits on the manager-liveness guard without touching the registrar. In the hypothetical where the manager did survive, re-adopting is the only correct response, since the caller needs a live closer.

### Coverage

Not examined in this slice: `authorization.ex`, `token_client.ex`, `discovery.ex`, `loopback_listener.ex`, and the store's mutation-lease internals beyond the fence reached from `refresh/3`. `authority.ex`'s new `valid?/1` was reviewed only for whether it can block or mutate under concurrency (it cannot — it is a pure rebuild-and-compare), not for round-trip completeness against every `from_host/4` field.

PR #1173, opened 2026-08-03
feat(kernel): promote model-authored source into an attested component

### LINKED ISSUE #1167: Promote model-authored source into an attested component
## Context

While dogfooding the incident-evidence compiler (`docs/plans/future/single-pass-compiler-evidence.md`), a question came up that the runtime half-answers: can a workflow have the model *write* the program it then runs, and can anything useful survive the run?

Probing the mission environment directly settled the first half — this all works today:

- source handed to `kernel/eval-source` may contain `def`/`defn`, and definitions **persist into later `eval-source` calls for the life of the run**;
- a dynamically defined function may call mission capabilities;
- it may **not** shadow a protected namespace's public exports — both `defn` and `def` against `incident.evidence/resolve-citations` are refused with `invalid_form: … it is a public export of the protected namespace …`, and the real function still answers afterwards.

So a model can author a working library inside a run. What it cannot do is leave one behind. A runtime `defn` is not in the `FrozenBundle`: not covered by the component source hash, not versioned, dead at end of run, and absent from `mission_inventory` — which is built once at run construction, so the model never sees its own library in its tool context, and with no declared `:signature` there is no `export-meta` for `fit/handles?` to read.

That gap is correct by design and should stay. `mission_inventory_hash` is load-bearing: it is frozen into `run-started` (`run_config.ex`) and folded into analysis-profile identity beside `bundle_hash` (`analysis_profile.ex`), so one value answers "exactly what surface was this model shown, for this whole run". `FrozenBundle` defends its own immutability with an in-VM attestation. Neither should be traded away for convenience.

The right path is promotion: generated source becomes a real component, compiled and attested, used by a *later* run. This issue is about closing that loop.

## What already exists

Most of the mechanism is built, which is why this is worth filing as a bounded piece of work rather than a design epic.

- **In-memory components.** `Component.new/1` takes `:source` as a binary with a bounded `:origin` (≤1024 bytes) "used for provenance and diagnostics", defaulting to `"memory"`. `Kernel.compile_bundle/1` takes `[Component.t()]`, not paths. Nothing requires generated source to be a file to be compiled.
- **Candidate evaluation, with host authority.** `PtcRunner.Kernel.ComponentOverride` plus `mix ptc.run --component-override-descriptor` already does the dangerous part properly. The descriptor carries exactly `component_id`, `base_source_hash`, `source_hash`, `path`. Verification is two-sided: the candidate hash proves the operator is compiling the bytes they think they extracted, and the base hash proves those bytes were derived from the component installed *now*, so a candidate written against a since-changed base is rejected rather than evaluated against the wrong baseline. Source is confined to the descriptor's directory, opened once, and the bytes hashed are the bytes compiled. The verified identity is recorded in `run-started` so a trial artifact names base *and* candidate. A manifest cannot request one and a generated program cannot observe one.
- **A compile-and-assemble gate with no execution.** `--check` passes through `run_options/1`, so `mix ptc.run MANIFEST --component-override-descriptor D --check` already answers "does this candidate compile and assemble" without running anything or spending a model call.
- **Source capture.** `RunBuilder.capture_prelude_sources/4` writes bundle sources into the `--inspect` artifact, so a run's exact component sources are already recoverable.

The `ComponentOverride` moduledoc states the boundary this issue stops at, and it is the right one:

> Nothing here writes. Materializing a candidate is a separate trusted host step, and promotion stays an explicit human decision.

## What is missing

**1. Materialization.** Nothing turns model-authored source into `{candidate.clj, descriptor.json}`. Computing `source_hash`, resolving the installed component's `base_source_hash`, and writing both into a confined directory is manual today. This is the trusted host step the moduledoc names and deliberately excludes — it just does not exist yet.

**2. Authoring provenance.** `origin` is a bounded free-form string and the descriptor has no field for it. A promoted component should be traceable to the run, model, and prompt that produced it; otherwise a candidate is anonymous source with a hash. `origin` is the natural carrier (1024 bytes is enough for a run id, model id, and prompt hash) but nothing defines a convention, and the descriptor does not round-trip it.

**3. A promotion gate.** What must a candidate satisfy before a human is asked to look at it? Compiles and assembles (`--check` covers this); declares a `:signature` and docstring for anything `:prompt`-visible, since without them it is invisible to `export-meta` and `fit/handles?`; does not widen effect or capability requirements relative to its base. The last one matters most — an override "changes which source compiles, never what compilation permits", but a candidate that reaches more capabilities than its base is a different risk profile and should be surfaced explicitly rather than merely permitted.

**4. Component-shaping.** A runtime `defn` is not component-shaped. Something has to wrap authored forms in an `ns` with visibility, a namespace docstring, and per-export `:signature`. Whether that is the authoring prompt's job, a post-processing step, or a template is open.

## The asymmetry worth deciding first

`ComponentOverride` replaces an **already-selected** component — `:override_component_not_selected` is one of its errors, and `base_source_hash` presumes a base exists. A generated helper library that did not previously exist has no base to derive from, so the existing mechanism covers *improving* a component but not *introducing* one.

Options, in rough order of preference:

1. Scope promotion to improving an existing component. A manifest that wants a generated helper declares a placeholder component first; the model rewrites it. Preserves both hash checks, no kernel change.
2. Allow a descriptor with a null `base_source_hash` for a genuinely new component id. Weakens the derived-from guarantee to nothing in that one case; would need a distinct code path and its own justification.
3. Leave new-component creation entirely to the ordinary human workflow, and let promotion cover only substitution.

Option 1 is likely right and costs nothing, but the choice belongs in review rather than in this description.

## Non-goals

- Making `mission_inventory` mutable or re-hashable mid-run. Breaks single-hash run identity, breaks analysis-profile identity, risks a mid-run ceiling breach, and breaks the run/`ReplSession` parity the module guarantees.
- Letting a manifest or a generated program name an override. That boundary is already correct and should stay.
- Automatic promotion. Promotion stays an explicit human decision; this issue is about making the decision cheap to reach and well-evidenced, not about removing it.

## Why it is worth doing

Without it, every model-authored program is disposable, and "the model wrote something better than the hand-written version" is unfalsifiable past the end of a run. With it, a generated component gets the same treatment as any other: hashed, attested, replayable, visible in the inventory, judgeable by `fit/handles?` — and a trial artifact that names both the base and the candidate, which is exactly what deciding between them needs.

Related: #1165, #1166 (both found in the same dogfooding pass).

andreasronge:
Closes #1167.

A model can already author a working library inside a run — `kernel/eval-source`
accepts `def`/`defn` and those definitions persist for the life of the run. What
it cannot do is leave one behind: a runtime `defn` is not in the `FrozenBundle`,
is not covered by a component source hash, and is absent from `mission_inventory`.

That gap is correct and stays. This closes the other loop: generated source
becomes a real component, compiled and attested, used by a *later* run.
`mission_inventory` is untouched.

## What lands

- **`mix ptc.materialize`** — the trusted host step `ComponentOverride`'s
  moduledoc names and deliberately excludes. Turns authored source into
  `{candidate.clj, descriptor.json}` and gates it.
- **`CandidatePromotion`** — G1 compiles · G2 prompt-visible exports declare a
  signature and docstring · G3 no export widens relative to its base · G4
  dependencies preserved by construction.
- **`CandidateArtifact`** — publication reusing the existing private-directory
  primitives.
- **Descriptor provenance** — an optional closed `provenance` object, plus the
  resolved override environment, now reach `run-started`.

## Design decisions worth reviewing

**Materialize first, then gate.** Override selection and environment resolution
are private (`application_package.ex:342`); the only public operation that
applies an override is acquisition, and it needs a materialized descriptor. So
the task publishes first and re-acquires through the descriptor it just wrote —
the exact path a run takes. The gate cannot see a package a run could not
produce.

**G3 compares per export ref, never an application-wide union.** If base export
`A` requires `read-tool` and `B` requires `write-tool`, then `A` gaining
`write-tool` adds no name outside the union while materially widening `A`. There
is a test for exactly this.

**The gate does not check capability grants, and says so.** Real capability names
exist only after provider acquisition. `ProviderDescriptor.provides` is
inter-provider *service atoms* disjoint from `requires`, not capability names,
and a dynamic provider discovers its tool names at acquisition — comparing them
would compare an atom against an unrelated string. A candidate naming an
ungranted capability passes here and fails at run-time assembly, which stays
authoritative. Documented, not hidden.

**Provenance is operator-asserted and kept out of content identity.**
`identity/1`'s output is the override content record payload, so putting an
asserted timestamp or acceptance flag there would make
`application_content_digest` depend on claims about authorship rather than on
content. A separate `attribution/1` carries it. There is no `model_id` field: a
descriptor is a published artifact, raw model selectors must not be published,
and hashing does not help because model ids are a small enumerable domain.

**A maintainer trap is now documented.** `ApplicationPackage` strips the override
environment in two places that look identical. `:375` feeds the package
projection (changed here). `:493` feeds a content record whose *name* already
encodes it — changing that would shift `application_content_digest` for every
override-bearing application. A regression test pins the digest.

## Verification

- `mix precommit` clean; credo reports no issues.
- Pre-push gates passed (root tests + `mix prepush`).
- 17 new tests. The end-to-end one is the claim under test: a placeholder stubs
  `double` returning `0`, authored source is materialized and gated, and a second
  run through the descriptor returns `42` — only reachable if the model-authored
  bytes compiled and executed.

## Notes

The plan (`docs/plans/lisp-kernel/promote-authored-component.md`) went through
five rounds of independent review, which killed 19 findings including several
that would have shipped broken — a gate built on `run --check` (deleted by the
accepted stable-CLI grammar), a capability check with no valid static source, and
a rename-based publication that POSIX allows to replace an empty directory.

Implementation then contradicted the plan once: an export-free placeholder cannot
work, because a consumer must depend on it and call something, so the base
application would not compile and there would be no baseline. A placeholder must
stub every export its consumers call. Corrected in the plan, the guide, and the
task docs.

Scope note: the issue lists four missing pieces and calls the non-widening
criterion the one that "matters most", so all four are here rather than
materialization alone.

https://claude.ai/code/session_017udL6wWtkfRs5DEReFpaLy


andreasronge:
## Merge-gate review outcome

Ran an independent review over the branch diff. Six findings, all legitimate; all fixed in `af8bd83d`.

**[P1] Stale semantic build projection.** `CandidatePromotion` is under the `lib/ptc_runner` semantic root and `priv/semantic_build_projection.json` was not regenerated, so `mix ptc.gen_semantic_revision --check` failed and CI would have gone red. Regenerated.

Worth flagging for anyone reading this later: my local `mix precommit` exited 0 and did **not** catch this. The semantic-revision check is not part of that gate.

**[P2] The gate could approve an unrunnable candidate.** If the promoted component owns the manifest entry, a candidate could drop or rename that function, compile cleanly — G3 would only note a removed export — and pass. Every later run would then fail in `RunCoordinator.validate_entry/2`. That defeated the gate's entire purpose in exactly the case where the promoted component is the entry owner. G1 now checks the same `PreparedRun.entry_callable?/2` the coordinator does.

**[P2] Unbounded reads.** `File.read/1` loaded the whole file before the 1 MiB limit was enforced, so a large input could exhaust the VM instead of returning the documented error — and the `--from-result` path used a stat-then-read pair that races a growing file. Both now open once and read at most the limit plus one byte.

**[P2] Incomplete RFC 6901.** Array indices (`/value/0/source`) and the root pointer (`""`) were rejected, so source under a list or at a document root was unreachable. Both supported.

**[P3] Doc contradiction.** The guide still called the descriptor an "exact four-field schema" two lines below a five-field example. Corrected.

Three regression tests added: entry-drop refusal, both pointer forms, oversized file.

After the fixes: `mix precommit` clean, `mix ptc.gen_semantic_revision --check` current, credo no issues, 20 tests on the new modules.

Issue #1166, opened 2026-08-03
Rejections that discard what the runtime already knows (CLI, language, prelude compiler)


andreasronge:
## Context

`docs/plans/lisp-kernel/analysis-query-ergonomics.md` item 1 fixed this for **trace queries**: a rejected query now names the offending key and the accepted set instead of collapsing to `invalid trace query`. The same defect exists in several other surfaces. Each was hit while dogfooding an application against the runtime; each cost a round trip that the runtime had the information to prevent.

## 1. `split` rejects a string separator with a bare type error

```clojure
(split "a-VERDICT-b" "VERDICT")   ;=> error
```

The signature is `(split s re-or-char)`. Passing a string — the obvious thing to try, and what `clojure.string/split` accepts in the form most people remember — produces a type error that does not say a regex or char was expected. Hit twice in one session, once inside an analysis session where the error surfaced only as `type_error`.

## 2. A private function named `parse` is reported as undefined

```clojure
(ns t) (defn- parse [x] x) (defn run [_] (return (parse 1)))
;=> "prelude body references undefined variable(s): parse"

(ns t) (defn- parse-report [x] x) (defn run [_] (return (parse-report 1)))
;=> compiles
```

Cause: `parse` is in `PtcRunner.Lisp.BuiltinNames.java_member_atoms/0`, from Java interop names such as `LocalDate/parse` and `Instant/parse`. There is no `name: "parse"` in `priv/functions.exs`, so nothing in the documented surface suggests the name is taken.

Two problems. The message blames the user's own definition as *undefined* when the real cause is a collision with a reserved member atom. And it is not obvious that unqualified Java member names should be reserved at all — `parse` is only ever reachable as `LocalDate/parse`.

## 3. `--inspect` path rejection does not state the convention

```
mix ptc.run … --inspect DIR/run.jsonl
** (Mix) ptc.run failed: {:inspection_preflight_failed, :invalid_inspection_path}
```

The artifact must be named `*.inspection.jsonl`. The error does not say so, and the requirement is easy to miss in the guide.

## 4. `--load` silently ignores a `return`

`mix ptc.repl --load file.clj` evaluates the file as setup and does not print. A `(return …)` at the end of the loaded file is silently discarded; the result expression must be given separately with `-e`. Nothing says so at the point of use.

## Not a bug, recorded so it is not "fixed"

[...]

PR #1174, opened 2026-08-04
fix(kernel): carry the diagnostics the runtime already computed

### LINKED ISSUE #1172: Private analysis sessions redact the analyst's own identifiers, and two smaller diagnostics that hide their cause
## Context

Analyzing an application's own private inspection records with the
`inspection-analysis-v2` profile — the intended workflow for
`mix ptc.repl --profile inspection-analysis-v2 --private-unattended`. Three
things cost round trips, and the first is a real defect rather than ergonomics.

Same family as #1166: in each case the runtime knows exactly what went wrong and
returns something that does not carry it.

## 1. A private session redacts identifiers that are the analyst's own script

An `unbound_var` failure inside the analysis profile answers only:

```
{"kind": "unbound_var", "message": "private evaluation failed",
 "reason": null, "capability_failure": null}
```

The identical fault outside a private session names the problem exactly:

```clojure
(kernel/eval-source "(defn- g [x] (* x 3)) (return (g 14))")
;=> {:kind :unbound_var :message "Undefined variables: defn-, g, x"}

(kernel/check-source "(defn- g [x] (* x 3)) (return (g 14))")
;=> {:outcome :invalid :diagnostic {:kind :unbound_var
;                                   :message "Undefined variables: defn-, g, x"}}
```

**The redacted names are not private data.** `defn-`, `g` and `x` come from the
analyst's own source text, typed into `-e` or a script file moments earlier.
The private resources are the *captured records* the session reads; the query is
the operator's own input, and it is already echoed back in the session's
`input_kind: "eval"` event.

Redacting it removes the one fact that fixes the script. Finding the cause here
meant checking every function in the expression against
`docs/function-reference.md`, then bisecting by elimination — two extra round
trips to recover a string the runtime had already computed.

There is a real boundary to respect: an unbound-variable message could in
principle carry a name derived from captured data. But the current rule
discards the common case (a typo in the operator's own query) to cover a rare
one, and does it silently. Options worth weighing: report names that appear
verbatim in the submitted source, keep the fault kind and drop only names that
do not, or keep the redaction but say *why* it is redacted.

## 2. `defn-` in dynamic source fails as an undefined variable

Not a private-session problem — the message above is from a normal evaluation:

```clojure
(defn f [x] (* x 2))  (f 21)   ;=> 42
(defn- g [x] (* x 3)) (g 14)   ;=> Undefined variables: defn-, g, x
```

`defn-` is a prelude/component definition form; in dynamic source the whole
form parses as a call to an undefined `defn-`, which is why `g` never binds.
That is defensible, but the diagnostic describes the symptom rather than the
cause. Someone porting a component helper into a REPL expression — an obvious
thing to do while debugging that component — reads "Undefined variables:
defn-, g, x" and has to work out that the *first* name is the actual problem
and the other two are consequences.

Saying `defn-` is not available outside component source would end it
immediately. Cheap either way; combined with #1 it was genuinely hard.

## 3. Profile resource directories read as empty rather than unmatched

`inspection-analysis-v2` takes `--resource traces=DIR --resource inspection=DIR`
and expects the artifacts flat inside them. Pointing it at a directory whose
artifacts sit one level down — which is what `run.sh` in the incident-compiler
experiments produces, one subdirectory per run tag — succeeds and answers:

```clojure
(inspection/runs {})
;=> {"items" [] "next_cursor" nil "omitted_count" 0 "truncated" false
;    "snapshot_hash" "sha256:b3014b0a…"}
```

A capture that matched nothing is indistinguishable from a capture that
genuinely holds no runs, and the session reports `ok` either way. This is
exactly the shape finding 4 in
`docs/plans/lisp-kernel/agent-developer-findings.md` describes — a rejected
query reading as an empty result — on the resource-loading path rather than the
query path. The session already computes a `snapshot_hash` over what it
captured, so it knows how many files it admitted; reporting zero admitted files,
or refusing a resource directory containing no artifacts at its own level,
would both close it.

## Why these together

Each one is small. Together they made a fifteen-minute analysis take an hour,
and all three share #1166's shape: the runtime holds the answer and returns
something that does not carry it. The private-analysis path is where it hurts
most, because it is the surface an operator reaches for precisely when
something has already gone wrong.

Related: #1166, #1165.

andreasronge:
Closes #1172.

Three defects on the private analysis path, all with the shape #1166 named: the
runtime holds the answer and returns something that does not carry it.

## 1. A private session no longer redacts the analyst's own identifiers

A private session must not forward evaluator message text — that text can quote
a captured record, and a host may route diagnostics somewhere the result never
goes. It can still say what the *operator's own source* already says.

So the diagnostic is rebuilt rather than forwarded. `PtcRunner.Kernel.PrivateDiagnostic`
admits a message only for a fault kind carrying structured, source-derived
detail (today `:unbound_var`), and every name must appear verbatim in the
submitted source. Every other byte is a literal in that module, so the
invariant is mechanical: a private diagnostic contains nothing but fixed text
and substrings of the operator's own input.

```clojure
(defn- g [x] (* x 3)) (return (g 14))
;=> Error (unbound_var): Undefined variables: defn-, g, x. Hint: 'defn-'
;   defines a private helper in component source only; use defn in dynamic source
```

Everything else keeps a fixed message and now says that it is fixed:

```clojure
("some-string" 1)
;=> Error (not_callable): private evaluation failed; diagnostic withheld by the
;   private result policy
```

Every error map the session emits carries `message_redacted?`, including the
separate result-limit constructor, so a withheld diagnostic is distinguishable
from one that was never produced. The canonical analysis trace is unchanged —
it never carried the message and still does not.

## 2. `defn-` and `ns` name their own cause

In dynamic source the whole form parses as a call to an undefined head, so the
head is the cause and the names it would have bound are consequences. The
diagnostic now says which is which, and carries the names structurally in
`details.unbound_names` — `Lisp.validate/2` reads that instead of parsing the
prose, and `kernel/check-source` diagnostics gain the same structured names.

## 3. A resource directory that matched nothing is refused

Directory capture is one level deep by design. A `--resource` directory whose
artifacts sit in subdirectories admitted zero files, succeeded, and answered
every query with an empty page — indistinguishable from a capture that holds no
runs. Both profiles now fail closed:

```
ptc.repl profile setup failed: the traces resource directory contains no
*.jsonl trace files at its own level; artifacts in subdirectories are not
captured
```

The refusal lives in profile capture and reuses each profile's existing
snapshot-cleanup path; snapshot primitives, `TraceLog`, and host-installation
acquisition keep their contracts. `TraceSnapshot` info gains `file_count`
(parity with `InspectionSnapshot`), and a started session reports what it
admitted, so a *near*-miss is visible too:

```
Captured traces: 12 files, 12 runs
```

With `--format jsonl` the same counts appear under `capture` in the
`session-started` record.

## Verification

`mix precommit` — root 5489 tests, Viewer 72, launcher 36 — plus
`MIX_ENV=test mix dialyzer`. Plan and review notes in
`docs/plans/lisp-kernel/private-diagnostics-and-capture-visibility.md`.

Reviewed by `codex` twice: once on the plan (one [P1] — the result-limit error
constructor bypasses `error_projection/2` and needed the flag set explicitly)
and once on the implementation (no [P1]; three [P2]s, all fixed: an improper
`unbound_names` list could raise instead of failing closed, a partially rebuilt
message did not say so in its own text, and three claimed regression tests were
missing). The second round independently traced the boundary: `unbound_names`
has exactly one producer, and canonical events never receive the projected
error message.


andreasronge:
Follow-up filed as #1175: pre-execution diagnostics (`invalid_arity`, `invalid_form`, `parse_error`, …) are still withheld in a private session even though the analyzer runs before any capability call. That needs a second admission rule with its own safety argument, not an extension of this PR's verbatim-in-source check — see the issue for why wiring `:invalid_arity` into the existing allowlist would deliver nothing.

PR #1222, opened 2026-08-09
fix: restore PR #1218 and #1217's CI step, and close #1220

### LINKED ISSUE #1220: --private-unattended --format jsonl with no input opens an interactive REPL, corrupting the JSONL stream
Shipped in PR #1218. Found by independent codex review after the branch was pushed; the fix was written and verified but did not make it into the merge, so `main` carries the defect.

## Reproduce

```bash
mix ptc.repl --profile inspection-analysis-v2 \
  --resource traces=<dir> --resource inspection=<dir> \
  --session-trace-dir <dir> \
  --private-unattended --format jsonl
```

Note: no `-e`, no `--load`, no script, no stdin.

**Expected:** rejected with `--format jsonl requires non-interactive profile input`, exactly as `log-analysis-v2` behaves.

**Actual:** a `session-started` JSONL record is emitted, then the human REPL banner (`PTC-Lisp REPL [inspection-analysis-v2] (Ctrl+D to exit; :help for commands)` and a `ptc> ` prompt) is printed into the same stream, then a `session-closed` JSONL record. The stream is not valid JSONL, and a non-interactive caller blocks on a prompt it never asked for.

## Cause

`AnalysisProfileRegistry`'s `@unattended_input_modes` includes `:interactive`:

```elixir
@unattended_input_modes [:interactive, :eval, :load, :script, :stdin]
```

`:interactive` means "wait for a human at a keyboard", which is the one thing `--private-unattended` rules out. Admitting it lets the call reach the interactive loop.

There is a second, related gap: `validate_profile_combinations/6` guards the "jsonl requires input" case with `:jsonl in recipe.frontend().output_formats`. That is the profile's *static* declaration, which for `inspection-analysis-v2` does not include `:jsonl` — `--private-unattended` widens it dynamically. So the guard that should have caught this was skipped.

## Fix

Two changes, both small:

1. Drop `:interactive` from `@unattended_input_modes`.
2. Make the jsonl-requires-input guard ask whether jsonl is reachable *for this call* (`:jsonl in output_formats or private_unattended?`), not just for the profile in general.

A third ordering detail falls out: once (1) lands, supplying both `--private-terminal` and `--private-unattended` reports `unsupported_profile_input` instead of `private_destination_conflict`, because input-mode authorization now runs first. Rejecting the destination conflict with a narrow early guard — before input/output authorization, leaving the single-destination paths' ordering untouched — restores the precise error.

I have this fix implemented and verified locally (`mix precommit` green: 5896 root / 72 Viewer / 37 launcher), including a regression test at both the registry and CLI layers. Happy to open the PR.

andreasronge:
## What happened

`77613b63` ("docs(cli): replace the standalone wrapper with a named envelope destination") reverted three merged pieces of work. Its own commit message says *"Plan and documentation only; no production code"* — which is how it passed review. It removed **224 lines** of production code, tests, and CI configuration.

Cause is confirmed: the branch was based on `5d59fe20`, predating those merges, and squashed with `git reset --soft origin/main`, which moves the branch pointer while keeping the older tree.

**Nothing failed, because the regression tests were removed in the same commit.** Main's own plan still described `--private-unattended` as shipped at `stable-cli-contract.md:872` while `git grep private_unattended origin/main -- lib/ test/` returned zero hits.

## Restored

Each verified as a pure reversion — `git show 77613b63 --numstat` on these paths is deletions-only or deletion-plus-rejoined-line, so restoring from `d2687a6f` loses nothing that commit added:

| From | Files |
|---|---|
| **#1218** | `analysis_profile_registry.ex`, `analysis_session_builder.ex`, `ptc.repl.ex`, both test files, `kernel-repl.md`, `AGENTS.md` |
| **#1217** | `.github/actions/setup-elixir/action.yml` — the bundled native launcher build step, gone entirely (5 references → 0) |

`priv/semantic_build_projection.json` is **regenerated, not restored** — it must describe this tree, which now also carries the fix below. It's checked only by the release gate, so a stale projection would have surfaced first at a release tag, claiming a build identity that doesn't describe its own sources.

## Closes #1220

`@unattended_input_modes` admitted `:interactive` — waiting on a human at a keyboard, the one thing "unattended" rules out. So `--private-unattended --format jsonl` with no `-e`/`--load`/script/stdin fell through to the interactive REPL loop and printed its prompt banner into a stream the caller relies on being valid JSONL.

The guard that should have caught it was skipped for a related reason: it asked `:jsonl in frontend.output_formats` — the profile's *static* declaration, which doesn't contain `:jsonl` for `inspection-analysis-v2` because `--private-unattended` widens it dynamically.

Rather than re-derive that in the frontend, the registry now answers it. `reachable_frontend/2` is the single source of what one invocation can reach, used by authorization, by the frontend guard, and by the profile description — which previously reported only the attended path, telling a caller half its own contract:

```json
"private_unattended": {
  "input_modes": ["eval","load","script","stdin"],
  "output_formats": ["clojure","jsonl"]
}
```

Requesting both destinations is now rejected by a **named** guard running before input/output authorization. Without it the conflict surfaces as `unsupported_profile_input` once `:interactive` is no longer admitted. It's a narrow separate check rather than a reordering, so single-destination paths keep their existing "which mode is wrong" ordering — and it's named so Checkpoint F's per-command declaration can point at one mutual-exclusion rule instead of restating it, which `stable-cli-contract.md` F3 commits it to carrying.

## Guard against a repeat

The restored **registry- and builder-level** tests are the protection. They sit outside the frontend extraction F3 plans, so they fail loudly if `ptc.repl.ex` is rewritten again without them. Restoring only the Mix-task tests would have left the same hole.

## Test plan

- [x] `mix precommit` green — 5896 root, 72 Viewer, 37 launcher
- [x] `MIX_ENV=dev mix docs --warnings-as-errors` clean
- [x] #1220 case now rejected: `--format jsonl requires non-interactive profile input`
- [x] `--private-unattended -e '(count (get (inspection/runs {}) "items"))'` returns real private inspection records end to end
- [x] Restored CI action carries #1217's launcher build step again (5 references)
- [x] Base verified current against `origin/main` immediately before push

### DIFF
--- a/.github/actions/setup-elixir/action.yml
+++ b/.github/actions/setup-elixir/action.yml
@@ -41,6 +41,21 @@ runs:
       working-directory: ${{ inputs.project-directory }}
       shell: bash
 
+    # The bundled launcher is the only dependency carrying a native artifact,
+    # and nothing in the Mix or make staleness chain ties that artifact to
+    # `c_src`. A restored `_build` keeps the cached executable, and a rebuilt
+    # one is downloaded from the release matching the launcher version. Either
+    # way the root project can run against an executable older than the tree
+    # it is testing, so drop the artifact and build this tree's source.
+    - name: Rebuild the bundled native launcher from source
+      if: inputs.project-directory == '.'
+      env:
+        PTC_RUNNER_LAUNCHER_BUILD_FROM_SOURCE: '1'
+      run: |
+        rm -rf "_build/${MIX_ENV:-dev}/lib/ptc_runner_launcher"
+        mix deps.compile ptc_runner_launcher
+      shell: bash
+
     - name: Cache Dialyzer PLT
       if: inputs.cache-plt == 'true'
       uses: actions/cache@v5
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,6 +11,12 @@ documentation guidance in `docs/guides/documentation-guidelines.md`, language
 reference in `docs/ptc-lisp-specification.md`, and built-ins in
 `docs/function-reference.md`.
 
+To debug the runtime itself (not a manifest under it) — query canonical
+traces or private inspection records (model exchanges, generated source,
+capability payloads) non-interactively — use `mix ptc.repl --profile
+inspection-analysis-v2 --private-unattended`. See ["Private analysis without a
+terminal"](docs/guides/kernel-repl.md#private-analysis-without-a-terminal).
+
 ## Working Style
 
 This is a **0.x library** — expect breaking changes. Backward compatibility is
--- a/docs/guides/kernel-repl.md
+++ b/docs/guides/kernel-repl.md
@@ -306,6 +306,44 @@ source and exact trace-query payloads are not copied into canonical events.
 The output directory cannot be the input directory, an ancestor or descendant
 of it, or the same physical directory through symlinked parents.
 
+### Private analysis without a terminal
+
+`inspection-analysis-v2` requires one authorized private destination. Two are
+available, and exactly one may be given:
+
+- `--private-terminal` — the attached terminal is the private sink. Requires
+  stdin and stdout to be terminals, and admits only interactive input.
+- `--private-unattended` — this command's own streams are the private sink.
+  Admits `-e`, `--load`, script, and stdin input, and `--format jsonl`.
+
+```bash
+mix ptc.repl --profile inspection-analysis-v2 \
+  --resource traces=tmp/traces --resource inspection=tmp/inspection \
+  --session-trace-dir tmp/analysis \
+  --private-unattended --format jsonl \
+  -e '(inspection/runs {})'
+```
+
+The attached-terminal check is an **accident guard, not access control**. It
+stops private values reaching a log or transcript by mistake; it cannot stop a
+determined caller, because `isatty` cannot tell a human's terminal from a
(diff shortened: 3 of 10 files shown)


andreasronge:
**Added a fourth restore** that the original PR missed: `test/mix/tasks/ptc_run_downstream_test.exs`, from `9ec7d0fd` ("test(mix): accept cold build output"). Caught by the session that authored the revert, reviewing against the pre-revert tree.

It matters more than its size suggests. The restored assertion takes the last line of stdout; the reverted one decodes the whole of it:

```elixir
# restored                                    # reverted
output                                        Jason.decode!(output)
|> String.split("\n", trim: true)
|> List.last()
|> Jason.decode!()
```

The downstream consumer test builds a project in a temporary directory. On a **cold** build the command's stdout carries compilation lines ahead of the envelope, so decoding all of it raises, while taking the last line succeeds. With a warm `_build` there is no compiler output and both forms pass — I verified this by running the reverted assertion locally, and it passes. It is also tagged `:slow`.

So the revert of the one commit that exists specifically to handle cold output was invisible to every local gate — `mix precommit`, the pre-commit hook, and the pre-push hook all went green — and would have failed only in CI. Worth recording as the reason a green local run was not sufficient evidence here.

Verified every reverted path is now either byte-identical to `d2687a6f` or differs only by the intentional #1220 fix.

Issue #1221, opened 2026-08-09
Exceeding an installed limit ceiling is reported as schema_violation, naming a schema the manifest satisfies


andreasronge:
Same shape as #1166 — the runtime knows the real reason and the rejection discards it.

## Reproduce

Any manifest whose `limits` exceed the `LimitCatalog` defaults, run **without** `--host-config`:

```bash
mix ptc.run app/ptc.json      # limits: {"mission_capability_calls": 512}
```

```json
{"error":{"code":"schema_violation",
          "message":"the application manifest does not satisfy its schema",
          "path":"/limits/mission_capability_calls",
          "phase":"application"}}
```

## Why the message is wrong

The manifest **does** satisfy its schema. `priv/schemas/ptc-application-manifest.schema.json` declares:

```json
"mission_capability_calls": {"type":"integer","minimum":1,"maximum":2592000000}
```

512 is comfortably inside that range. The actual constraint is the installed ceiling in `lib/ptc_runner/kernel/limit_catalog.ex`:

```elixir
{:mission_capability_calls, 128, 128},   # {field, default, ceiling}
```

so the value is refused for exceeding **128** — a number that appears nowhere in the diagnostic, and a rule that lives nowhere in the schema the message names.

Confirmed by bisection: `{"mission_capability_calls": 128}` passes, `512` fails; the same holds for `workflow_capability_calls` (ceiling 64) and every other catalogued limit.

## Why it is costly

The remedy the message implies — "fix the manifest to match the schema" — is a dead end, because the manifest already matches it. The actual remedies are invisible from the diagnostic: lower the value to the ceiling, or install a higher ceiling with `--host-config`. A host config raising `mission_capability_calls` to 8192 makes the identical manifest run, which is the thing worth telling the caller.

This cost real time to diagnose on a manifest that was correct and had run fine before the catalog ceilings tightened.

## Suggested

Report it in its own right — the limit name, the requested value, the installed ceiling, and that a host configuration can raise it — rather than as a schema violation. `mission_capability_calls: 512 exceeds the installed ceiling 128` is self-remedying; the current message actively misleads.

Worth checking whether the phase is right too: this is a limits/installation constraint, not a manifest-syntax one.

andreasronge:
Filed #1266, which determines how much of this issue is actually achievable now.

Re-verified the mechanism at `d71c32d9`, and the loss happens one layer earlier than the projection catch-all:

[...]

PR #1270, opened 2026-08-10
feat(kernel): give the envelope span field a producer

### LINKED ISSUE #1266: Envelope detail carriers: span is fully built with zero producers, notes is pinned to const []
## Summary

Several open diagnostics issues (#1246, #1221, #1249, and the #1166 family) all want a rejection to say *where* it happened or *what value* broke *what bound*. The command envelope has two fields for exactly that. Neither is usable today, but for completely different reasons and at completely different cost:

- **`span` is fully built, validated, schema-permitted — and has zero producers.** Populating it needs no contract change at all.
- **`notes` is nailed shut**: typed as the literal `[]`, hardcoded in two places, and pinned to `{"const": []}` in the published V1 schema. Populating it is a contract decision.

Filing this separately because it is a prerequisite that changes the shape of the other fixes. Doing #1246/#1221 first produces slightly better fixed strings and then requires revisiting the same rows to add the actual detail.

## `span` — built and unused

The machinery is complete in `lib/ptc_runner/kernel/command_diagnostic.ex`:

```elixir
:31   @type span :: %{start_byte: non_neg_integer(), end_byte: non_neg_integer()} | nil
:51   allowed = [:source, :path, :span, :subject, :provider_activity]
:66   true <- valid_span?(span, source),
:264  defp valid_span?(nil, _source), do: true
:266  defp valid_span?(%{start_byte: start_byte, end_byte: end_byte} = span,
                       %CommandSource{byte_size: byte_size}), do: ...
:142  defp span_map(%{start_byte: start_byte, end_byte: end_byte}), ...
```

`new/3` accepts it, `valid_span?/2` bounds-checks it against the source's `byte_size`, `span_map/1` serialises it.

The published V1 schema **already permits a real span object** on 32 diagnostic rows, including every row these issues care about:

```
application/schema_violation      bundle/compile_failed        execution/workflow_failed
application/reference_missing     bundle/entry_invalid         result_cleanup/result_invalid
application/input_invalid         bundle/bundle_invalid        execution/mission_failed
… 32 total; the remaining 87 rows pin span to {"type": "null"}
```

And no caller anywhere in `lib/` passes `:span`:

```console
$ grep -rn "span:" lib/ --include="*.ex" | grep -v "form_scanner\|events.ex\|command_diagnostic.ex"
$ # (empty)
```

So every envelope ever emitted has `"span": null`. Verified on real runs — a `bundle/compile_failed` envelope:

```json
{"code":"compile_failed","message":"the component bundle could not be compiled",
 "source":{"kind":"component","name":"main.clj"},"span":null,"notes":[],"path":null}
```

The byte offsets are not hypothetical: `PtcRunner.Lisp.Prelude.FormScanner` already produces byte-exact top-level form spans (`lib/ptc_runner/lisp/prelude/form_scanner.ex:140-149`, `@type byte_span :: {non_neg_integer(), non_neg_integer()}`).

**Cost: producer-side only.** No catalog row, no schema regeneration, no contract version question, no consumer breakage — the schema already allows what would be emitted.

## `notes` — actually closed

```elixir
:41   notes: [],                    # the type is the literal empty list
:80   notes: [],                    # hardcoded at construction
:125  "notes" => [],                # hardcoded again in the JSON projection
```

and in `priv/schemas/ptc-command-envelope-v1.schema.json` there is exactly one distinct `notes` schema across the whole document:

```json
{"const": []}
```

So a consumer validating against V1 **rejects** any envelope with a non-empty `notes`. This is the field that would carry "requested 512, ceiling 128", and opening it is a V1 compatibility decision, not a refactor.

## What this blocks

`source`, `path`, and `subject` are the only live carriers. Consequence: a diagnostic can say *which document* and *which JSON pointer*, but never *which byte range* and never *what value violated what bound*.

- **#1246** (compile errors carry no position) — the location half is unblocked by the `span` work alone, with no contract change. The message half is separate.
- **#1221** (limit ceiling reported as `schema_violation`) — a better code and message are achievable now, but "you asked for 512, the installed ceiling is 128" needs `notes`. Both numbers are bound and discarded at `lib/ptc_runner/kernel/manifest.ex:818-829`.
- **#1249**, **#1166** — same shape: the codes can improve now; the values cannot.

**Not blocked by this:** #1176. That rejection travels the Dispatcher/agent-loop feedback channel, not the command envelope, so it needs its own fix and should not wait on this.

## Suggested split

1. **`span` producers** — populate it wherever a diagnostic has a source and a known byte range, starting with `bundle/compile_failed` from the form scanner. Self-contained, no contract risk, immediately useful to #1246.
2. **`notes` decision** — decide whether V1 may gain populated notes (relaxing `{"const": []}` to a bounded, closed-shape array is arguably additive for tolerant consumers but breaks strict ones), or whether that waits for a V1.1 envelope. Worth answering before #1221 is scheduled, because the answer determines whether #1221 is a code rename or a real fix.

Piece 1 is worth doing regardless of how piece 2 is decided.

andreasronge:
Closes #1266 (piece 1 in full, and answers piece 2).

## What this changes

The command envelope's `span` was fully built, validated, and schema-permitted with **zero producers** — every envelope ever emitted said `"span": null`. Component compilation now populates it.

- **`Prelude.Compiler`** tags a failure with the index of the top-level form it was walking. This happens once, in the walk itself, rather than at each of the ~30 error sites.
- **`Prelude.ErrorSpan`** (new) resolves that index — or a `ref`/`namespace`, for failures raised after the walk — to a byte range through `FormScanner`. The scanner cross-checks its own form list against the real parser head-by-head and refuses to return on any disagreement, which is what makes indexing by position safe rather than a guess.
- **`BundleCompiler`** forwards the range on `:component_compile_error`, bounded against the component's own bytes.
- **`RunCoordinator`** binds `bundle/compile_failed` provenance to those exact bytes with `CommandSource.with_bytes/3`, which is what lets `CommandDiagnostic`'s existing bounds check admit the span at all.

Resolution fails **open**: a source that will not scan, an out-of-range index, or an ambiguous name keeps the null span. A missing span is the status quo; a wrong one points a reader at innocent code.

No catalog row, schema, or contract version changes — the published V1 schema already permitted a span object on these rows, so `mix ptc.gen_docs --check` is clean.

## The `notes` half

Answered in the other direction, and recorded in `docs/guides/kernel-maintainer.md` and `CommandDiagnostic`'s moduledoc: V1 pins `notes` to `{"const": []}`, so a populated array invalidates the envelope for every strict V1 consumer. Reporting a rejected value against the bound it broke needs a later envelope version, not a producer. That settles the scheduling question #1266 raised for #1221 — the code and message there can improve now; the two numbers cannot travel until V1.1.

## Verification

- New compiler tests assert the **slice**, never the offsets: cutting the source at the resolved span must yield the offending form.
- A `validate` envelope carries a span that slices to `(defn broken)`, and stays schema-valid; an unlocatable parse failure keeps `"span": null`.
- `mix precommit` green end to end.

## Review

One `codex-independent-review` round (xhigh) found two ways an in-bounds span could still blame innocent code — an ambiguous ref (`dep_ref_in_def` is raised *before* duplicate names are rejected) and `Enum.at/2`'s negative-index reverse lookup. Both fixed with regression tests in the second commit; the follow-up round is clean.

### DIFF
--- a/docs/guides/kernel-maintainer.md
+++ b/docs/guides/kernel-maintainer.md
@@ -605,6 +605,18 @@ retain their workflow or mission occurrence.
 A non-null byte span is admitted only when its
 `CommandSource` was constructed with the exact trusted source bytes and the
 exclusive end offset is within that retained byte bound.
+Component compilation is the one producer of that span. When the prelude
+compiler can attribute a failure to a single top-level form,
+`PtcRunner.Lisp.Prelude.ErrorSpan` resolves that form's byte range from the raw
+component text, and `bundle/compile_failed` carries it against provenance bound
+to the same bytes. Failures no single form owns — parse errors, whole-bundle
+limits — keep the null span. Spans are best-effort by design: an unresolvable
+one is dropped rather than guessed, because a wrong range points a reader at
+innocent code.
+`notes` stays pinned to the empty list. The published V1 schema fixes it as
+`{"const": []}`, so any populated array makes an envelope invalid for a strict
+V1 consumer; a diagnostic that has to report a rejected value against its bound
+needs a later envelope version, not a relaxed V1.
 The command-specific host loader preserves closed phase-2 causes instead of
 projecting every failure to `host_invalid`: inaccessible files are
 `host_unavailable`, invalid bytes/JSON are `host_invalid`, structural failures
--- a/lib/ptc_runner/kernel/bundle_compiler.ex
+++ b/lib/ptc_runner/kernel/bundle_compiler.ex
@@ -13,6 +13,7 @@ defmodule PtcRunner.Kernel.BundleCompiler do
   alias PtcRunner.Kernel.FrozenBundle
   alias PtcRunner.Lisp.Parser
   alias PtcRunner.Lisp.Prelude.Compiler
+  alias PtcRunner.Lisp.Prelude.ValidationError
 
   @bundle_hash_domain <<"ptc.frozen-bundle.v2", 0>>
   @max_components 128
@@ -228,7 +229,8 @@ defmodule PtcRunner.Kernel.BundleCompiler do
             %{
               reason: :component_compile_error,
               id: component.id,
-              details: error_message(error)
+              details: error_message(error),
+              span: failure_span(error, component.source)
             }}}
       end
     end)
@@ -344,4 +346,15 @@ defmodule PtcRunner.Kernel.BundleCompiler do
     do: String.slice(message, 0, 4_096)
 
   defp error_message(error), do: inspect(error, limit: 10, printable_limit: 4_096)
+
+  # The byte range of the top-level form the compiler blamed, in the exclusive
+  # end-offset form command diagnostics carry. Bounded against the component's
+  # own source so a diagnostic can never claim a range the source cannot hold;
+  # anything unlocatable stays `nil`, exactly as before.
+  defp failure_span(%ValidationError{span: {offset, length}}, source)
+       when is_binary(source) and offset >= 0 and length >= 0 and
+              offset + length <= byte_size(source),
+       do: %{start_byte: offset, end_byte: offset + length}
+
+  defp failure_span(_error, _source), do: nil
 end
--- a/lib/ptc_runner/kernel/command_diagnostic.ex
+++ b/lib/ptc_runner/kernel/command_diagnostic.ex
@@ -6,6 +6,11 @@ defmodule PtcRunner.Kernel.CommandDiagnostic do
   Contract-authorized paths must also match the sealed contract authority bound
   to their source classification. Rendering never inspects a lower-level
   reason or rejected value.
+
+  `notes` is reserved and always empty: the published V1 envelope schema pins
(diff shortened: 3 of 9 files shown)


andreasronge:
Superseded by #1269, which landed this work on `main` as `ebc162db` ("feat(kernel): give the envelope span field a producer") — the same commit, including both fixes from the codex review round, with `lib/ptc_runner/lisp/prelude/error_span.ex` byte-identical to this branch. #1269 then went further on top of it: `compile_reason` classification into the `syntax_invalid` / `undefined_variable` / `duplicate_definition` rows, span resolution moved into its own bounded worker so optional attribution cannot reclassify a compile error, and additional tests.

Rebasing this branch onto current `main` leaves nothing:

```
Rebasing (2/2) dropping d16a88c2 fix(lisp): never resolve a span through an ambiguous ref -- patch contents already upstream
$ git log --oneline origin/main..HEAD   # (empty)
$ git diff origin/main HEAD --stat      # (empty)
```

Every conflict hunk was this branch reverting #1269's newer work, so there is nothing to resolve. Closing.

PR #1276, opened 2026-08-11
fix(kernel): explain rejected capability arguments

### LINKED ISSUE #1176: A rejected capability argument names neither the argument nor the bound it violated
## What happens

When a model-authored capability call violates the installed tool's
`inputSchema`, `Dispatcher` refuses it with:

```
kind=:protocol_error; reason=:invalid_arguments
```

and the feedback the agent loop returns to the model is:

> The capability call failed without an unsafe effect. kind=:protocol_error;
> reason=:invalid_arguments. Send one corrected run_ptc_lisp call with corrected
> capability arguments.

That names no argument, no constraint, and no bound. If the model cannot infer
which argument was wrong, it has nothing to correct, and "send one corrected
call" is an instruction it cannot act on.

## Why it matters — it cost a whole benchmark arm

Found while running the incident-evidence compiler's agent-loop arm against a
332-record evidence corpus. The tool is an MCP `evidence_search` whose schema
caps `limit` at 50:

```elixir
"limit" => %{"type" => "integer", "minimum" => 1, "maximum" => 50}
```

The model read `matched: 332` from a search result, correctly concluded it was
holding a subset, and asked for a bigger page:

```clojure
(incident.evidence/search "schema-migration-stall" nil nil 100)
```

Refused. It re-sent the **identical call 18 times** and the run died on the
`protocol_errors` budget having fetched zero records. Its reasoning was right at
every step; it simply had no way to learn that 50 was the ceiling.

Critically, **the declared `:signature` does not carry the bound either**:

```
(incident_id :string, query :string?, source :string?, limit :int?) -> {...}
```

`limit :int?` is all the model ever sees. So there is no surface — signature,
description, or error — from which the maximum is discoverable. The only way to
find it is to guess numbers until one is accepted.

## Why this is safe to say

The argument name and its constraint both come from the **operator's own
declared `inputSchema`** in `ptc-host.json`. Neither is captured data, neither
originates from a provider response, and the rejection happens *before* the
capability executes. This is the same verbatim rule that #1174 established for
`unbound_var` (names verified against the operator's own submitted source) and
that #1175 argues for pre-execution analyzer diagnostics.

A message of the form

> `limit` must be <= 50 (got 100)

leaks nothing that `ptc-host.json` does not already state.

## Suggested shape

Report the failing property path and the violated keyword from the JSON Schema
validation that already runs, e.g. `{:argument "limit", :constraint "maximum",
:limit 50}`. `JSONSchema.valid?/2` currently discards this — `JSV.validate/3`
returns the specific violation before it is reduced to a boolean.

A cheaper partial fix that would still have unblocked the case above: surface
only the argument *names* that failed validation, without values.

## Related

- #1166 — rejections that discard what the runtime already knows (CLI, language,
  prelude compiler). Same theme, different surface.
- #1175 — pre-execution diagnostics withheld in private sessions.
- #1172 / #1174 — the verbatim-naming precedent.

Evidence and the full failure trace are recorded in
`docs/plans/future/single-pass-compiler-evidence.md` on branch
`worktree-incident-evidence-compiler`.

andreasronge:
Closes #1176.

## What changed

- retain up to three bounded, schema-authored argument violations when a capability input schema rejects a call
- surface the declared argument path, violated keyword, and small declared bound or expected set to the shipped agent correction loop
- keep submitted values, undeclared property names, provider details, and custom semantic-validator reasons opaque
- bound explanation work by input-node and raw-error budgets, falling back to separately bounded boolean validation for large inputs
- document the correction-feedback contract and add dispatcher plus end-to-end agent-loop coverage

## Root cause

Dispatcher previously collapsed every input-schema rejection to `invalid_arguments`, and the agent feedback prelude rendered only the error kind and reason. The model therefore could not tell which schema-declared argument or bound needed correction.

## Validation

- test-first regression coverage (initial focused assertions failed before implementation)
- independent cumulative Codex review; all actionable findings addressed
- `mix test` focused kernel/agent suite: 186 passed
- `mix dialyzer`: 0 errors
- `mix precommit`: duplication baseline 62/current 62/new 0; 6,018 root tests, 72 Viewer tests, and 40 launcher tests passed; package and standalone-release checks passed
- pre-push hook with `PTC_PRE_PUSH_MAX_CASES=2`: all checks passed in 219 seconds

### DIFF
--- a/docs/guides/building-agents.md
+++ b/docs/guides/building-agents.md
@@ -507,8 +507,11 @@ that evaluator-owned provenance separately. Rebuilding, copying, or
 destructuring an equal error-shaped map after a read does not qualify. The
 shipped PTC-Lisp loop then gives the model one correction turn. An error-shaped
 value after an unrelated read is still terminal, as is any failure after a
-`write` or undeclared effect. Correction feedback includes only the bounded
-`kind` and `reason` classification, never provider details.
+`write` or undeclared effect. Correction feedback always includes the bounded
+`kind` and `reason` classification. An input-schema rejection may additionally
+name up to three schema-declared argument paths, violated keywords, and small
+declared bounds or expected sets. Submitted values, undeclared property names,
+opaque semantic-validator reasons, and provider details remain withheld.
 
 When a failure is genuinely unsafe to retry, the shipped loop still does not
 repeat the program — but it no longer discards the run either. It spends one
--- a/docs/guides/building-agents.md
+++ b/docs/guides/building-agents.md
@@ -510,8 +510,19 @@ value after an unrelated read is still terminal, as is any failure after a
 `write` or undeclared effect. Correction feedback always includes the bounded
 `kind` and `reason` classification. An input-schema rejection may additionally
 name up to three schema-declared argument paths, violated keywords, and small
-declared bounds or expected sets. Submitted values, undeclared property names,
-opaque semantic-validator reasons, and provider details remain withheld.
+declared numeric, string-length, or item-count bounds. Enum and const literals,
+submitted values, undeclared property names, opaque semantic-validator reasons,
+and provider details remain withheld. These bounded details are part of the
+ordinary `tool/...` error map, so plain PTC-Lisp workflows can inspect them too;
+they are not added to the canonical trace.
+
+If the bounded host validator times out, exhausts its heap, crashes, or returns
+an unrecognized result, the call instead returns
+`capability_unavailable/input_validation_unavailable`. This is not evidence
+that the model's arguments violate the schema: it charges neither a protocol
+error nor a capability call. The shipped agent loop uses evaluator-owned call
+provenance to fail the outer workflow immediately, without asking the model for
+a correction or spending another provider turn.
 
 When a failure is genuinely unsafe to retry, the shipped loop still does not
 repeat the program — but it no longer discards the run either. It spends one
--- a/docs/guides/kernel-maintainer.md
+++ b/docs/guides/kernel-maintainer.md
@@ -1155,16 +1155,31 @@ sandbox for malicious BEAM code.
 When a model-authored call fails the capability's compiled input schema before
 callback entry, Dispatcher retains at most three correction facts for the agent
 loop: the schema-declared argument path, the violated keyword, and a small
-declared bound or expected set when one fits. Paths are resolved through the
-frozen schema, the root is `$`, array positions are rendered as `[]`, and
+declared numeric, string-length, or item-count bound when one fits. Enum and
+const literals are always omitted. Paths are resolved through the frozen
+schema, the root is `$`, array positions are rendered as `[]`, and
 non-identifier names use JSON-quoted bracket notation so punctuation and
-controls cannot alter path structure. Explanations are omitted when the
-submitted structure or raw error set exceeds its fixed work budget; large
-structures use a separately bounded boolean validation whose timeout and heap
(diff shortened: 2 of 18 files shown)


andreasronge:
Reviewed `codex/issue-1176-schema-validation-details` @ `eecf68c` by running probes against the branch locally (Elixir 1.20.2 / OTP 29.0.3), not by reading alone.

## Confirmed working as described

Issue #1176's exact case now produces:

```
The capability call failed without an unsafe effect. kind=protocol_error;
reason=invalid_arguments; violations=limit violates maximum 50. Send one
corrected run_ptc_lisp call with corrected capability arguments.
```

`50` present, submitted `100` absent. The `details` list also reaches **plain PTC-Lisp with no facade and no agent loop** — a bare `(return (tool/evidence_search {"limit" 100}))` gets `%{"details" => [%{"argument" => "limit", "constraint" => "maximum", "expected" => 50}]}` — so custom workflows can branch on it, not just `agent.feedback`. Worth stating in the guide; the current wording reads as if this is agent-loop-only.

## No canonical-event disclosure

A rejected call emits no capability event at all — only `protocol_errors: 1` in usage. Violations reach the model and Lisp, never the public trace. That matches the intent and is worth keeping a test on.

## Two things worth settling before this leaves draft

**1. Enum members cross a facade boundary.** `project_expected/1` echoes declared enums up to 8 members / 256 bytes. With `"tier" => %{"enum" => ["internal-alpha", "partner-beta"]}` the model receives:

```elixir
%{"argument" => "tier", "constraint" => "enum",
  "expected" => ["internal-alpha", "partner-beta"]}
```

This is operator-declared, so it satisfies the rule as written. But that rule was framed around direct capabilities, and a *complete* prompt-visible facade exists precisely to narrow what the model sees. The rejection path bypasses that narrowing. If any schema's enum carries internal tiers, environment names, or partner identifiers, one bad call surfaces them. Bounded rather than unbounded, but it deserves an explicit decision and a line in the guide rather than being left implied.

**2. Large valid arguments can now fail closed.** Arguments over `@max_explanation_nodes` move from an inline `valid?` to `BoundedWorker.run/2` with a timeout and heap cap, and any exhaustion maps to `{:error, []}` → `invalid_arguments`. The suite covers the happy path (175k rows with raised bounds) but not a worker that times out under scheduler pressure, where a previously-valid call would now be refused. A contention test would close this; it is the only risk here I would want resolved before merge.

Lower-priority: `project_violation/2` matches JSV's internal error shape (`:kind`, `{:properties, name}`, `:items`). A library upgrade could stop producing violations — the tests would fail loudly rather than degrade silently, so this is a note, not an objection.

## Scope note

This makes the bound learnable *after* one failure; the issue's own re-measurement puts the residual cost at one turn rather than zero.

The bound is already rendered for direct capabilities — `model_contract.ex:200` emits `minimum`/`maximum` — so the real gap is that a facade signature drops what the raw tool entry carries. Carrying bounds through the facade signature (or not dropping them when the facade suppresses the raw entry) would take the cost to zero turns, because the model never writes `100` in the first place. That is the fix that prevents the failure rather than explaining it, and it is cheap next to 664 lines. Not a blocker for this PR — but if only one lands, that one is worth more.

Caveat on my own evidence: I verified the rendering behaviour against a minimal facade, where the raw `tool/` entry is still listed alongside the facade entry and the bound stays visible. The 18-retry failure requires a facade complete enough to suppress those entries. Worth confirming that precondition against the incident-evidence manifest before treating the diagnosis as settled.

---

PR #1378, opened 2026-08-14
fix(cli): report provider local doctor failures

### LINKED ISSUE #1372: friction(doctor): the provider/<alias>/local check aborts for a missing MCP binary and falsely passes for a missing replay fixture
## Summary

`doctor` publishes a `provider/<alias>/local` check whose job is to answer "is this provider's
local environment usable?". It is wrong in **both** directions, and the two failures are opposite:

- **Missing MCP binary → the whole command aborts.** No checks at all, no `readiness` field —
  `doctor` dies with the same `local_preflight/environment_unavailable` that `run` would give.
- **Missing replay fixture → the check reports `pass`.** `doctor` says
  `{"code":"available","name":"provider/frozen-model/local","status":"pass"}` for a fixture file
  that does not exist on disk, while `run` on the identical inputs fails preflight on that exact
  check name.

This is the third input class in the `#1302`/`#1354` family. `#1354` fixed the invalid-manifest
class (verified fixed this pass — `doctor` now reports `application: fail/schema_violation`).
The mechanism here is the same: `doctor` does not own the local check, so it either inherits
preflight's abort or never consults it.

## Repro

From a fresh checkout.

**Direction A — abort instead of a failed check.** Point the shipped filesystem sample at a
command that does not exist:

```console
$ python3 - <<'PY'
import json
h = json.load(open('examples/kernel-tutorial/ptc-host.json'))
h['install']['workspace']['transport']['command'] = 'definitely-not-a-real-binary-xyz'
json.dump(h, open('/tmp/host-badmcp.json','w'), indent=2)
PY
$ mix ptc doctor examples/kernel-tutorial/03-file-agent/ptc.json --host-config /tmp/host-badmcp.json
** (Mix) error: local_preflight/environment_unavailable: provider/workspace/local at mission[0]: a required local environment is unavailable (run_ref: …)
```

No `checks` array, no `readiness`. `doctor` is the command you reach for *because* a provider is
broken; it refuses to run for the reason you invoked it.

**Direction B — false pass.** A replay alias whose fixture file is absent:

```console
$ mkdir -p /tmp/replaywalk && cd /tmp/replaywalk
$ cat > agent.clj <<'EOF'
(ns eval.agent "Replay evaluation walk." {:visibility :prompt})

(defn run [input]
  (agent.core/run (get input "task") {"max_turns" 2}))
EOF
$ cat > ptc-host.json <<'EOF'
{"install": {"frozen-model": {"source": "llm_replay",
  "installation_revision": "frozen-model-v1", "fixtures": "replay.jsonl"}}}
EOF
$ cat > ptc.json <<'EOF'
{
  "version": 1,
  "workflow": {
    "components": [
      {"id": "eval.agent", "path": "agent.clj", "dependencies": ["agent.core"]},
      {"library": "agent.core"}
    ],
    "entry": "eval.agent/run"
  },
  "input": {"value": {"task": "Return the number 42."}},
  "providers": {"workflow": [{"name": "frozen-model"}]},
  "labels": {"name": "replay-walk", "tags": {"mode": "agent"}},
  "missions": {"default": {"components": []}}
}
EOF
$ # note: replay.jsonl is deliberately NOT created
$ cd /path/to/ptc_runner
$ mix ptc doctor /tmp/replaywalk/ptc.json --host-config /tmp/replaywalk/ptc-host.json
{"checks":[…,{"code":"available","name":"provider/frozen-model/local","status":"pass"},
 {"code":"declarative","name":"provider/frozen-model/selection","status":"pass"}],
 "readiness":"unverified"}

$ mix ptc run /tmp/replaywalk/ptc.json --host-config /tmp/replaywalk/ptc-host.json
** (Mix) error: local_preflight/environment_unavailable: provider/frozen-model/local at workflow[0]: a required local environment is unavailable (run_ref: …)
```

`doctor`: `provider/frozen-model/local` **pass**. `run`: `provider/frozen-model/local` **fails
preflight**. Same alias, same host document, same working directory, same check name, opposite
answers. An empty `replay.jsonl` and a malformed one behave identically.

## Observed

`doctor` cannot report a failing local environment. It aborts for MCP and false-passes for
`llm_replay`.

## Expected

Both classes produce a row:

```json
{"name": "provider/workspace/local",     "status": "fail", "code": "command_not_found"}
{"name": "provider/frozen-model/local",  "status": "fail", "code": "fixtures_unreadable"}
```

with `readiness: "failed"`, exactly as the `application` check now does after `#1354`.

## Why it hurts a newcomer

`doctor` is what the quickstart names when something goes wrong. Direction A removes the
diagnostic precisely when it is needed. Direction B is worse than removing it: a green `doctor`
is positive evidence that the setup is fine, so a newcomer will look everywhere except the file
that is actually missing.

## Severity

Blocker for the diagnose-my-setup path. Direction B (false pass) is the more damaging half.

## Bug or correct-but-frictional

**Bug**, in both directions. `readiness: "unverified"` does not excuse direction B: the row
asserts `status: "pass"`, not "unknown", for a condition the check is named after and that the
runtime evaluates a second later.

## Suggested fix

Give `doctor` its own bounded local probe per provider source rather than sharing preflight's
abort path: `stdio` → resolve the command on `PATH`; `llm_replay` → stat and parse the fixture
file, reporting entry count. Then run every check to completion and let `readiness` summarise,
which is the contract `#1354` already established for `application`.

## Environment

Elixir 1.20.2 / OTP 29.0.3, macOS 25.3.0, commit `cbf9ced0`, `mix ptc` from a source checkout.
Model-independent — no model is acquired on either path.

andreasronge:
## Summary

- collect every bounded provider-local finding in plain doctor and report failed rows instead of aborting
- validate `llm_replay` fixtures with the same parser and ceilings used by acquisition before reporting availability
- add closed `command_not_found` and `fixtures_unreadable` diagnostics, with matching docs and generated schema updates

## Root cause

Plain doctor delegated to runtime's fail-fast local preflight, while replay descriptors declared no local check. Missing stdio executables therefore escaped as bare command errors, and missing replay fixtures were reported as available even though `run` rejected them at local preflight.

## Validation

- pre-push gate: 6,202 core tests passed
- static analysis, spec examples, generated artifacts, duplication gate, and Dialyzer passed
- Viewer: 44 tests passed
- launcher package and standalone release verification passed
- `MIX_ENV=dev mix docs --warnings-as-errors` passed
- focused regressions cover missing commands, missing and valid replay fixtures, and multiple simultaneous local failures

Closes #1372


andreasronge:
Addressed the follow-up review findings in f1f7dd3e (after merging current main):

- raised the audited-local worker heap ceiling so valid near-limit replay fixtures pass doctor preflight
- split missing MCP commands (`command_not_found`) from present-but-unusable executables (`executable_unavailable`)
- corrected the RunCoordinator docs to describe both the fail-fast and findings entries
- fixed an additional bounded-worker failure found during analysis: executable validation returned bare `false` for a non-executable file, crashing the worker and collapsing the result to `internal_error`
- regenerated the command-envelope schema and added end-to-end regressions

Validation: 245 focused tests, ExDoc with warnings as errors, `mix precommit`, and the complete pre-push gate (6210 core tests plus static analysis, Dialyzer, release, and Viewer checks). All passed.

andreasronge:
Final conflict resolution pushed in 6b324859 after `main` advanced again. The generated command-envelope schema was regenerated from the merged catalogs (preserving both the replay request-hash work and `executable_unavailable`), and the merged replay/doctor integration set passed 268 tests. The complete post-merge pre-push gate also passed: 6246 core tests, static analysis, Dialyzer, release, Viewer, and launcher validation.

Issue #1503, opened 2026-08-18
smoke: end-user pass 2026-08-18b (llama-4-scout / meta)


andreasronge:
# smoke: end-user pass 2026-08-18b (llama-4-scout / meta)

Second pass on 2026-08-18, run against `9e8c0c64` — **after** the fix sweep that closed nine of the
ten findings from `#1483` (PRs #1484/#1485/#1492/#1493/#1494). The subject is again someone who
downloaded `ptc-0.13.0-macos-arm64.tar.gz` and has nothing else.

## Run

| | |
| --- | --- |
| Date | 2026-08-18 (second pass; `#1483` was the first) |
| Commit | `9e8c0c64` — 14 commits after `0923ce89`. The sweep closing #1453, #1373, #1474–#1481, #1487, #1490, #1491. |
| Artifact | `scripts/package_standalone_release.sh` → `_build/artifacts/ptc-0.13.0-macos-arm64.tar.gz`, 108 s, 56 MB, extracted to `~/ptc-smoke-2026-08-18b/ptc`. `ptc --version` in 3.0 s from a cold extract. |
| Model | `openrouter:meta-llama/llama-4-scout`. **meta is a provider family no previous pass has used** (openai, google, deepseek, moonshot, qwen, xai, nvidia, z-ai, microsoft, mistralai, amazon). Uncataloged in `supported_models.json`; pricing resolved correctly throughout. |
| Toolchain | macOS 26.3.1 arm64, node v20.19.0, `@modelcontextprotocol/server-filesystem` 2026.7.10 |
| Branch | `smoke/llama4scout-2026-08-18b` |
| Wall clock | ~70 min after the artifact existed |
| Model spend | **$0.0505** — 211 model calls across 35 traced runs (`$0.04725` llama-4-scout, `$0.00327` deepseek from the shipped tutorial control) |
| Method | Orchestrator built the artifact, replayed the ledger, and filed. **Five** independent `claude -p` testers, cwd `~/ptc-smoke-2026-08-18b/*`, one per use case, launched concurrently, no shared session. |

**Pre-flight note on the model.** The very first tool-calling probe against `meta-llama/llama-4-scout`
returned an upstream `502` from Groq — *"tool call validation failed: parameters for tool addone did
not match schema"*. It did not recur in 211 subsequent calls. Recording it because a pass that hit it
once mid-walk would have blamed PtcRunner.

## Isolation check

Asked from the scratch directory before any task was given. Verbatim:

[...]

Issue #1501, opened 2026-08-18
friction(config): schema failures lack violation kinds; invalid projects bypass envelope publication


andreasronge:
## Maintainer triage — revised scope (2026-08-21)

The original report identified a real usability failure, but two parts have since been partially fixed and the literal suggested fix conflicts with the published diagnostic contract. The remaining work is scoped as follows.

### Already resolved

- Invalid project documents now report `arguments/project_invalid` instead of blaming the command line with `invalid_arguments`.
- Host validation now descends through tagged-union branch errors and retains safe closed-schema depth such as `/install/*/ceilings/timeout_ms`.

These improvements do not resolve the issue: project failures remain contentless and do not publish a requested envelope, while host and application-manifest failures still identify a location without identifying the violated rule.

### Revised objective

A recognized command that names an invalid hand-authored document should report a bounded, privacy-safe violation kind and the deepest schema-authorized path available. A valid invocation with an invalid project document should produce a normal diagnostic outcome and publish `--envelope`. True command-line syntax rejections remain envelope-free.

### Workstream 1 — shared schema-violation projection

Introduce one bounded representation carrying:

- a schema-authorized path; and
- a finite violation kind such as `unknown_property`, `required`, `type`, `minimum`, `maximum`, or `pattern`.

Apply the representation across host, application-manifest, and project validation where each decoder has the information. Render a bounded message that names the rule and, when present, the safe path.

Acceptance examples:

- `host configuration contains an unknown property at document root`
- `host configuration violates maximum at /install/*/ceilings/timeout_ms`
- wrong type and bad pattern at the same path produce distinguishable diagnostics

Do not populate `notes`: envelope V2 pins it to `[]`.

### Workstream 2 — project documents are document failures

Stop collapsing every `ProjectConfig` decoder failure to the undifferentiated `:project_invalid` reason before command admission can retain the diagnostic.

Give project-schema failures a document-owned phase/code, proposed as `project/project_schema_invalid`, with project source/path authority. Preserve structured decoder reasons for at least unknown property, missing required property, wrong type, range, pattern, and cross-field constraint failures.

A syntactically valid command naming an invalid project is not an argv rejection.

### Workstream 3 — envelope publication boundary

Retain and validate a requested `--envelope` destination before producing the project-schema diagnostic, then publish the ordinary failure envelope without bootstrapping providers or opening project references.

The boundary is intentional:

[...]

Issue #1500, opened 2026-08-18
friction(debug): `transcript/incomplete_evidence` is really an ambiguity count the runtime computed and threw away


andreasronge:
## Summary

`ptc transcript` refuses roughly a quarter of real agent runs with
`transcript/incomplete_evidence: transcript evidence is incomplete or ambiguous`. The condition is
almost always **ambiguity**, not incompleteness — `canonical_complete?` is `true` and
`missing_exchange_count` is `0` on every failing run I measured. The runtime has already computed
`ambiguity_count` and discards it.

Two independent testers hit this, and each derived a different wrong cause. Neither guess was close.

## Repro

Any project with `artifacts.trace: true` and `artifacts.inspection: true`; run an agent loop enough
times that some runs take many turns.

```console
ptc transcript RUN_ID \
  --traces .ptc/traces --inspection .ptc/inspection \
  --private-unattended --private-output ~/out/RUN_ID.json
```

Then, for a run that failed, start the Viewer with `viewer.private: true` and ask the ungated
endpoint the same question:

```console
curl -s http://127.0.0.1:PORT/api/analysis/runs/RUN_ID/conversation \
  | python3 -c 'import json,sys;c=json.load(sys.stdin)["conversation"];print({k:c[k] for k in ("complete?","canonical_complete?","ambiguous?","ambiguity_count","missing_exchange_count")})'
```

## Observed

Across 16 real runs of one agent project, 12 produced a transcript and 4 were refused. For the
refused ones:

```
cmd-7dkdbeh3sdt3mfw4zggtp3062p {'ambiguity_count': 5, 'ambiguous?': True,  'canonical_complete?': True, 'complete?': False, 'missing_exchange_count': 0}
cmd-2jq5w7kf2ztzfsggw5vrzqqqfy {'ambiguity_count': 2, 'ambiguous?': True,  'canonical_complete?': True, 'complete?': False, 'missing_exchange_count': 0}
```

and for one that succeeded:

```
cmd-7xxy0hgarqkk9ns9jf7hc901tm {'ambiguity_count': 0, 'ambiguous?': False, 'canonical_complete?': True, 'complete?': True,  'missing_exchange_count': 0}
```

Nothing is missing. The reconstruction is ambiguous — and the count is right there.

What it cost the two testers:

- One correlated failure against run length across 8 runs and concluded "runs with ≥13 LLM calls
  fail" (~6 min). Size correlates loosely but is not the rule: a 47-line artifact failed and another
[...]

Issue #1424, opened 2026-08-15
friction(cli): ptc init scaffolds a project that ptc models rejects as invalid_arguments


andreasronge:
## Summary

`ptc init` scaffolds a project document that `ptc models` then rejects as having bad arguments. Both
commands are in `ptc --help`; the argument form used is verbatim the one `ptc models --help` prints.
The real cause — the scaffolded project has no `host` block, and `ptc init` does not create one — is
never mentioned, and the error class blames the user's command line instead.

## Repro

```console
$ mix ptc init /tmp/newapp
created main.clj, ptc.json, ptc-project.json

$ mix ptc models --help
Usage:
  ptc models PROJECT.json | --host-config HOST.json

$ mix ptc models /tmp/newapp/ptc-project.json
** (Mix) error: arguments/invalid_arguments: use the documented arguments for this command
   (run_ref: cmd-4j8qxagw5agyxyb7kk988n2e2f)
```

The arguments *are* the documented arguments. Adding a `host` block to the scaffolded
`ptc-project.json` makes the identical command work:

```console
$ cp /path/to/a/ptc-host.json /tmp/newapp/ptc-host.json
$ python3 -c "import json; d=json.load(open('/tmp/newapp/ptc-project.json'));
  d['host']={'path':'ptc-host.json','env_file':{'path':'.env'}};
  json.dump(d,open('/tmp/newapp/ptc-project.json','w'),indent=2)"

$ mix ptc models /tmp/newapp/ptc-project.json
{"installations":[{"accepts_data":["normal"],"alias":"deepseek",…,"source":"llm"}]}
```

## Observed

- `ptc init` emits exactly one line — `created main.clj, ptc.json, ptc-project.json` — and no next
  step. The scaffold runs (`mix ptc run /tmp/newapp/ptc-project.json` prints `{}` in 1.6 s and
  correctly creates `.ptc/traces` and `.ptc/envelopes`; the new project launch config is genuinely
  nice), but it can never call a model, and nothing says how to change that.
- `mix ptc doctor /tmp/newapp/ptc-project.json --connect` on the host-less scaffold reports
  `readiness: "ready"`, `provider_activity: false`, `model_aliases: []`. Passing `--connect` and
  being told "ready" without a single request having been made is a claim a newcomer will over-read.
- `arguments/invalid_arguments: use the documented arguments for this command` is circular even when
  it is correct: it points at the documentation the user just followed. It never names which
  argument, or what about it was wrong.

## Expected

[...]

PR #1507, opened 2026-08-18
fix: make seven diagnostics name the rule they already computed

### LINKED ISSUE #1499: bug(host): a trailing newline in a credential reports internal_error, and `file:` credentials are the documented way to store secrets
## Summary

A credential whose value ends in a newline makes the command fail with `internal_error`
("the command failed internally"), the one diagnostic class that names nothing and implies a
PtcRunner defect rather than a user-fixable input. `printf '%s\n' "$TOKEN" > tok.txt` — or any
ordinary `echo`, editor save, or `gh auth token > file` — produces exactly that value, and the host
reference recommends `file:` credentials for secrets.

Isolated cleanly by an isolated tester with a 2×2.

## Repro

```console
# a credentialed HTTPS MCP installation, credential delivered by file
printf 'ghp_notarealtoken\n' > tok.txt          # trailing newline
cat > ptc-host.json <<'JSON'
{"credentials":{"tok":{"file":"tok.txt"}},
 "install":{"github":{"source":"mcp","installation_revision":"gh-v1",
  "transport":{"type":"streamable_http","endpoint":"https://api.githubcopilot.com/mcp/",
    "auth":[{"scheme":"bearer","binding":"tok"}]},
  "tools":{"get_me":{"as":"gh.me","effect":"read"}}}}}
JSON
ptc run proj.json

printf 'ghp_notarealtoken' > tok.txt            # no trailing newline
ptc run proj.json
```

## Observed

| credential | result |
| --- | --- |
| `file:` **with** trailing newline | `internal_error` — "the command failed internally", phase `internal` |
| `file:` without trailing newline | `authentication_rejected` ✅ |
| `literal:` with `\n` appended | `internal_error` |
| `literal:` clean | `authentication_rejected` ✅ |

The newline is the only variable.

## Expected

Either trim trailing whitespace when reading a credential (the common choice), or reject it with a
named, actionable diagnostic — `credential_invalid: the credential value contains a newline` — in
the `credential`/`host` phase rather than `internal`.

## Why it hurts a new user

`internal_error` is the one code that tells a user the fault is not theirs, so they stop looking at
their own inputs. The tester needed ~4 minutes and a controlled 2×2 to find it, and only because it
had already been burned by a different silent failure and had started A/B-ing everything. Writing a
token to a file with any normal tool adds the newline, and `ptc docs host` actively recommends
`file:` for secrets — so this is on the recommended path.

## Severity

**Time-sink**, and a likely-common one. Bug.

## Suggested fix

`String.trim/1` on the credential value after reading (both `file` and `literal`), or a
`credential_invalid` diagnostic naming the offending character class. Do not let it reach the
`internal` phase.

## Environment

`ptc-0.13.0-macos-arm64.tar.gz` (built from `9e8c0c64`), macOS 26.3.1 arm64.

This is a bug.

andreasronge:
Sweep A from the #1503 friction pass: seven diagnostic sites that had the
answer in hand and dropped it. One symptom, three very different costs, so
the batching follows the contract boundary rather than the symptom.

## The cost model, stated up front

The sweep was scoped on the assumption that "enrich `notes` without minting
codes" is the cheap path and that message text is expensive because it moves
`test/fixtures/command-human-v2.json`. Both are backwards:

- `notes` is `{"const": []}` in the published envelope schema, and
  `CommandDiagnostic`'s moduledoc says populating it "would invalidate the
  envelope for every strict V2 consumer... a later-version change, not a
  producer-side one." It is the one option that is closed.
- `command-human-v2.json` is a 37-line fixture read by one test. Not a cost.

The real gate is `priv/schemas/ptc-command-envelope-v2.schema.json`: 98
fully-enumerated branches with `message` pinned to `const` strings. So the
policy this PR follows, and the one worth keeping: **never populate `notes`;
prefer `path` depth; when text is genuinely required, mint a bounded message
the way the replay-fixture branch already does.** The one code minted here
adds a branch and removes zero lines.

## What changed

**Outside the closed catalog — free.** `ptc transcript` builds its own
free-form diagnostics, so #1500 cost one `case`. The projection measures three
conditions and `complete?` collapses them; every refusal measured in #1500 was
ambiguity (`canonical_complete?` true, `missing_exchange_count` 0) reported as
"incomplete", and two testers derived two different wrong causes. Each
condition now reports under its own code with its magnitude, and all three
point at the ungated Viewer route. The same command distinguished four
`--private-output` filesystem causes and rendered two near-identical messages
(three testers, three different inferred rules) — now one code and one named
rule each, the #1231 treatment `ptc run` already had.

The Viewer answered `inspection_not_configured` whenever the store was absent
and told the reader to set `artifacts.inspection` — which they had. Both that
and `viewer.private` must hold; the host states which is missing, and the
second needs no re-run.

**Inside the existing contract.** `#1501`'s host half: `/install` was the
producer's limit, not the schema's — the envelope already admits an 8192-char
pointer. Only the alias is caller-owned; `ceilings`, `transport`, and `tools`
beneath it are closed. A caller-chosen member is now elided rather than
truncated at, and only where the map's own `propertyNames` proves no
admissible name could render as the placeholder. `install/*/tools` carries
upstream tool names and admits `*` itself, so it still stops. Combined with
reading JSV's `oneOf` branch contexts (the union hides the broken member),
five of six host cases now carry distinct pointers, up from one shared by all
six.

**Two real bugs found on the way.** `--env-file` was gated on the selected
provider being an LLM, so an MCP-only project never read the file and then
failed with `credential_unavailable` naming a variable the operator had
supplied in it. And a credential ending in a newline reported
`internal/internal_error` — `AcquisitionReason` documents that
`:credential_unavailable` "cannot arrive on a command path"; the header
renderer's CR/LF check was the counterexample. Credentials trim, control
characters are refused at resolution where they are classified, and the
comment now names the invariant that keeps it true.

**Misclassification.** A project document failing its schema was reported as
`arguments/invalid_arguments`, sending the reader to `ptc help run` for a
switch mistake that was never there. Follows the `project_host_undeclared`
precedent from #1424.

**REPL.** A workflow session rejecting a mission's namespace answered with
thirty-odd language namespaces and never named `--mission`. The analyzer
cannot say it, and its text is reverse-parsed by
`NamespaceDiagnostic.rejected_namespace/1`, so it must not change. Only a
manifest session offers the hint: a direct session carries one synthetic
mission the read model cannot tell from a manifest declaring one, and
`--mission requires --manifest`.

## Deliberately not changed

- **`error.source.name` stays `ptc-host.json`.** It is `{"const":
  "ptc-host.json"}` in the envelope, and a user-authored path is exactly what a
  closed diagnostic refuses to echo. #1501 reads a contract as a defect.
- **A rejection still publishes no envelope.** True of every rejection, not
  this one, so making admission publish is its own decision and its own PR.

## Verification

Every fix has a failing test written first. Two tests caught things before the
code existed: `tools` admitting the placeholder, and — after extending the
walker-agreement property test to actually generate paths through open maps,
which it never had — a latent trap where a node carrying both `properties` and
`additionalProperties` would have the two walkers disagree and cost the
diagnostic its whole path.

`mix precommit` exits 0: 6680 root, 111 Viewer, 42 launcher; duplication
baseline=44 current=44 new=0; `mix docs --warnings-as-errors` clean. The
envelope schema diff is purely additive (288 lines added, 0 removed).

Closes #1499. Refs #1500, #1501, #1231, #1424.


andreasronge:
## Codex review — 3 findings, all [P2], no [P1]

Ran after the rebase onto `origin/main`. One was a real regression and is fixed in 11257831; the other two I checked and am declining, with reasons.

**[P2] Control-character rule too broad — valid, fixed.** Rejecting every control character at credential resolution was right for one sink and wrong for the other. An HTTP header cannot hold CR or LF; a child process environment entry legitimately can, and `transport.env` exists to hand a whole PEM block or JSON service-account key to an MCP server. That configuration worked on `main` and this branch would have broken it. `valid_secret?/1` is back to NUL-only; the header sink keeps its own rule and now answers `:mcp_authentication_failed`, which is the translation `AcquisitionReason` already prescribes for a rejected credential. #1499's `internal_error` is still closed — at the sink, rather than by a rule that overreached. New test asserts a multi-line credential survives resolution intact.

**[P2] `oneOf` branch ranked by depth can pick the wrong branch — tested, did not reproduce.** The specific case named was an `source: "llm"` install carrying `transport.start_timeout_ms`, predicted to report `/install/*/transport/start_timeout_ms` from the MCP branch. It actually reports `/install/*/transport`, which is the correct fault for an LLM install — that branch has no `transport` member. A branch whose `source` discriminator fails does not appear to contribute deeper errors. The heuristic is still a heuristic, and it is the one `command_validation_path/1` already used before this branch; the pointer can only ever name schema-owned members, never a value. Worth a discriminator-first selection if a real misdirection turns up — not on a case I could construct.

**[P2] A non-`ptc-project` `kind` now reports `project_invalid` — declining.** True, and it is the better answer. A document declaring `"kind": "ptc-manifest"` is not a valid project document, so the code is accurate. The suggested fix — emit this only for `"ptc-project"` documents — would send that case back to `arguments/invalid_arguments`, which is exactly the misdirection this PR removes.

`mix precommit` exits 0 after the fix: 6681 root, 111 Viewer, 42 launcher.

PR #1575, opened 2026-08-21
feat(llm): add opt-in PtcLlmHttp adapter

andreasronge:
Opt-in, end-user-testable `PtcLlmHttp` adapter. This is a trial seam, not the ReqLLM removal or default cutover. **Stay in draft** until ptc_llm_http#15 and #16 are in an exact published pin and all three coverage layers below are green against that pin.

## End-user configuration

```elixir
# mix.exs
{:ptc_llm_http, "== 0.1.0"}

# config/config.exs
config :ptc_runner, :llm_adapter, PtcRunner.LLM.PtcLlmHttpAdapter
```

```console
mix ptc run examples/kernel-tutorial/02-deepseek-extract.ptc-project.json
```

`PtcRunner.LLM.ReqLLMAdapter` remains the shipped default and the control adapter. There is no manifest adapter module and no ad hoc environment switch.

Do not report this Mix command green until deterministic loopback, live non-stream, and live streaming have all succeeded against the final exact dependency in a fresh OS process.

## What landed

- Exact optional `ptc_llm_http == 0.1.0` (`optional: true`, `runtime: false`). Ordinary builds compile and run without it.
- `PtcRunner.LLM` streaming is now a synchronous callback (`stream/3`). `ReqLLMAdapter` consumes its enumerable internally so observable stream and usage behavior stay the same.
- `PtcRunner.LLM.PtcLlmHttpAdapter` uses public PtcLlmHttp APIs and accessors. Error classification reads documented 0.1.0 struct keys because that package has no instance accessor. PtcRunner owns selectors, credentials, deadlines, budgets, retry refusal, usage projection, tracing redaction, and error classification.
- `PtcRunner.LLM.PtcLlmHttpRuntime` supervises the PtcLlmHttp Runtime as a permanent application child when `ptc_llm_http` is compiled. Capacity is released after success, error, callback failure, or caller death.
- Unsupported selectors are rejected during preparation. They never fall back to ReqLLM.
- Parallel coverage for the two live defects, without adapter workarounds or a 40M budget bump:
  - Minimized OpenRouter terminal-usage replay: finish event, the same empty index-zero terminal choice with usage, then `[DONE]`.
  - Fresh-BEAM public-hostname DNS probe (`https://example.com`). Hosts-file `localhost` is rejected as loopback before the DNS-role CA store load, so it cannot detect #16.
  - Live streaming OpenRouter E2E with ReqLLM as the non-stream control for stable result/usage shape.

## Supported / rejected selectors

Supported:

- `openrouter:model` → `https://openrouter.ai/api/v1`, public HTTPS, bearer credential
- `openai-compat:http://127.0.0.1:port/v1|model` → credential-free literal loopback HTTP
- `openai-compat:https://hostname/v1|model` → public HTTPS hostname

Rejected (no ReqLLM fallback): `ollama:`, `anthropic:`, `amazon_bedrock:`, `openai:`, empty OpenRouter, non-loopback HTTP, HTTPS IP literals, and credentialed HTTP loopback.

## Coverage layers (do not conflate)

| Layer | How | Status on published `0.1.0` |
| --- | --- | --- |
| Deterministic loopback | `mix test` adapter/smoke tests, including the OpenRouter terminal replay and fresh-BEAM `example.com` DNS probe | Green. Replay and DNS probe currently classify the published #15/#16 errors faithfully (`:invalid_result`); they must succeed after the pin bump. |
| Live non-stream | `mix test --include e2e` `llm/request` host installation | Previously succeeded after resolver warming. A **fresh** OS process still hits #16 (`resource_limit_exceeded` during DNS). |
| Live streaming | `mix test --include e2e` adapter `stream/3` | Added in this follow-up. On `0.1.0` it is expected to classify #15 (`:malformed_stream` → `:invalid_result`, `:possibly_dispatched`) after delivering deltas. |
| End-user `mix ptc run` | documented command, fresh OS process, `PtcLlmHttpAdapter` | **Not green.** Cold DNS exhausts the 4M aggregate (#16). Do not treat a warmed VM as evidence. |

ReqLLM remains the control: the live streaming case runs the same prompt through `ReqLLMAdapter.call/2` and compares the stable content/usage shape and credential redaction.

## Tests

- Deterministic Kernel `llm-request` loopback coverage (text, stream + usage, structured output, tools, credential refusal, timeout, disconnect/callback/caller-death cleanup and readmission, malformed tool arguments, quota 429)
- OpenRouter terminal-usage replay (finish + repeated empty terminal choice + `[DONE]`)
- Fresh-OS-process cold public-hostname DNS regression
- Optional `:e2e` OpenRouter host-installation **non-stream** and **stream** tests; skip when `OPENROUTER_API_KEY` is absent (ExUnit 1.20 skip tags)
- Package verification proves the Hex requirement is exact and optional, packaged source compiles without the dep (stub prepare fails), and compiles with the dep (real adapter prepares)

## Verification status

- Focused adapter/smoke/dep tests: **passed**
- `scripts/ci/core-tests.sh`: **passed** (6901 passed, 389 excluded)
- `mix precommit`: **passed**
- `MIX_ENV=dev mix docs --warnings-as-errors`: **passed**
- Live OpenRouter non-stream / stream: **not rerun here** (`OPENROUTER_API_KEY` absent)
- Documented `mix ptc run` with `PtcLlmHttpAdapter` in a fresh OS process: **not green** (blocked on #16)
- Adapter error mapping and `@default_process_budget_words` `4_000_000`: **unchanged**

## Join point after upstream #15 and #16

Once PtcLlmHttp publishes an exact patch:

1. Pin that version in `mix.exs` / `mix.lock` / package verification / docs.
2. Keep or document the aggregate process budget using the upstream-supported cold-hostname value. Do not treat a silent 40M bump as the fix.
3. The OpenRouter replay and fresh-BEAM DNS tests must take the success / post-DNS-error arms (no `:invalid_result` from #15/#16).
4. Rerun the documented Mix command in a fresh OS process.
5. Rerun both live E2E tests with `OPENROUTER_API_KEY`.
6. Prefer also landing ptc_llm_http#14 so Error field reads can move to the public accessor.

## Gaps before making PtcLlmHttp the default

- Host LLM installations always require a credential, and PtcLlmHttp forbids HTTP + credential, so deterministic host-installation success is live HTTPS OpenRouter only
- V1 streaming is text-only (tools/schema are non-stream `call/2`)
- Prompt cache is unsupported (`cache: true` is rejected)
- No tagged `PtcRunner.LLM.Target` host source yet
- Ollama / Anthropic native / Bedrock remain ReqLLM-only
- Usage guarantees, request limits, and catalog status are adapter-owned rather than host-declared
- PtcLlmHttp 0.1.0 Error values have no instance accessor; classification reads the documented struct keys
- Published 0.1.0 still has #15 (OpenRouter terminal usage) and #16 (cold DNS budget)

Do not merge, tag, or publish.

<div><a href="https://cursor.com/agents/bc-4e89a96e-b424-45e3-9cda-672343a66a8f?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-4e89a96e-b424-45e3-9cda-672343a66a8f&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `PtcRunner.LLM` streaming is now a synchronous callback (`stream/3`) that
+  receives normalized `%{delta: text}` maps and returns the final
+  `%{content: ..., tokens: ...}` response or a classified error. `ReqLLMAdapter`
+  consumes its enumerable internally so callers keep the same observable
+  stream and usage behavior.
 - `mix precommit` is nested fetch plus the quality scripts. The suite, Viewer,
   launcher package, and release verification run on `git push` (and in GitHub
   Actions), so an agent that already ran `mix precommit` should not follow it
@@ -16,9 +21,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Dev and test builds pin published `ptc_llm_http` `0.1.0` for loopback
-  streaming compatibility coverage. The package is not a production runtime
-  dependency and is not selected for ordinary requests.
+- Opt-in `PtcRunner.LLM.PtcLlmHttpAdapter` for downstream hosts that add exact
+  `{:ptc_llm_http, "== 0.1.0"}` and select the adapter with
+  `config :ptc_runner, :llm_adapter, PtcRunner.LLM.PtcLlmHttpAdapter`.
+  `PtcRunner.LLM.ReqLLMAdapter` remains the shipped default. Ordinary builds
+  compile and run without the optional dependency; selecting the adapter
+  without it fails during preparation and never falls back to ReqLLM.
+- Dev and test builds pin published `ptc_llm_http` `0.1.0` as an optional
+  runtime-selectable HTTP adapter and as loopback streaming compatibility
+  coverage. The package is not started by default and is not selected for
+  ordinary requests.
 - ptc-runner.dev now publishes the pinned MCP 2026-07-28 wire schema at
   `/schemas/mcp-2026-07-28.schema.json`, which is what a third-party author
   needs to write a server PtcRunner can acquire. It is upstream's document
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Opt-in `PtcRunner.LLM.PtcLlmHttpAdapter` for downstream hosts that add exact
   `{:ptc_llm_http, "== 0.1.0"}` and select the adapter with
   `config :ptc_runner, :llm_adapter, PtcRunner.LLM.PtcLlmHttpAdapter`.
-  `PtcRunner.LLM.ReqLLMAdapter` remains the shipped default. Ordinary builds
-  compile and run without the optional dependency; selecting the adapter
-  without it fails during preparation and never falls back to ReqLLM.
+  `PtcRunner.LLM.ReqLLMAdapter` remains the shipped default and the control
+  adapter. Ordinary builds compile and run without the optional dependency;
+  selecting the adapter without it fails during preparation and never falls
+  back to ReqLLM. Deterministic loopback coverage includes a minimized
+  OpenRouter terminal-usage replay and a fresh-BEAM cold-hostname DNS
+  regression. Do not treat the documented `mix ptc run` path as green until
+  deterministic, live non-stream, and live streaming coverage have all
+  succeeded against the exact published pin.
 - Dev and test builds pin published `ptc_llm_http` `0.1.0` as an optional
   runtime-selectable HTTP adapter and as loopback streaming compatibility
   coverage. The package is not started by default and is not selected for
--- a/docs/maintainers/development-setup.md
+++ b/docs/maintainers/development-setup.md
@@ -39,8 +39,11 @@ lives in `test/ptc_runner/llm/ptc_llm_http_adapter_test.exs` and includes:
 - deterministic loopback streams, including a minimized OpenRouter terminal
   usage replay (finish event, usage event with the same empty terminal choice,
   then `[DONE]`)
-- a fresh-OS-process cold-hostname regression (`https://localhost`) that must
-  not fail with a process-budget exhaustion during DNS
+- a fresh-OS-process cold-hostname regression against a public HTTPS
+  hostname (`https://example.com`) that must not fail with a process-budget
+  exhaustion during DNS once the published pin includes ptc_llm_http#16.
+  Hosts-file names such as `localhost` are rejected as loopback before that
+  DNS-role work runs, so they cannot detect the defect.
 
 Those two fixtures cover the published `0.1.0` defects tracked as
 [ptc_llm_http#15](https://github.com/andreasronge/ptc_llm_http/issues/15) and
@@ -52,8 +55,7 @@ succeeded against the final exact dependency:
 
 1. Deterministic loopback (always in `mix test`)
 2. Live non-streaming OpenRouter `llm/request` (`--include e2e`)
-3. Live streaming OpenRouter (`--include e2e`; the current live test was
-   non-streaming and cannot detect #15)
+3. Live streaming OpenRouter (`--include e2e`)
 
 To trial the adapter from a downstream host or this checkout:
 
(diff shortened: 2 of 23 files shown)


andreasronge (commented):
## Blocking live E2E findings and parallel fix plan

The first credentialed public-HTTPS run found two defects that the deterministic matrix cannot exercise. This PR should remain draft until both are fixed and the live path is rerun.

### Live evidence

1. **Cold DNS exhausts the configured process budget before dispatch.**

   The documented `mix ptc run` command fails in a fresh BEAM with the adapter's 4,000,000-word budget:

   ```elixir
   %{kind: :resource_limit_exceeded, phase: :dns, scope: :capacity, dispatch: :not_sent}
   ```

   Raising only `@default_process_budget_words` to 40,000,000 makes the same command complete with the same output as the ReqLLM control. The resolver is warm thereafter, so ordinary long-lived test VMs hide the problem.

   This is tracked upstream as [ptc_llm_http#16](https://github.com/andreasronge/ptc_llm_http/issues/16). PtcLlmHttp owns the opaque role partition: 4M gives DNS only 200K words, while 40M gives it 2M. Do **not** move DNS outside the bounded attempt and do not treat the 40M aggregate as the final fix without accounting for the tenfold increase to every other role ceiling.

2. **OpenRouter's real terminal usage event is rejected after all content has been delivered.**

   A live stream delivers every delta and correct content, then returns:

   ```elixir
   %{kind: :malformed_stream, phase: :stream, scope: :provider,
     dispatch: :possibly_sent, http_status: 200}
   ```

   A credential-free loopback replay of the captured bytes proves the trigger: OpenRouter's usage event repeats the empty index-zero terminal choice and `finish_reason` instead of sending `choices: []`. Changing only that event to `choices: []` succeeds with the same content and usage.

   This is tracked upstream as [ptc_llm_http#15](https://github.com/andreasronge/ptc_llm_http/issues/15). Do **not** special-case or salvage this in the PtcRunner adapter; the strict stream state machine must accept the narrowly bounded real wire form upstream.

### PtcRunner work that can proceed in parallel

Prepare these changes on this PR while #15 and #16 are implemented upstream:

- Add a minimized deterministic compatibility fixture for the observed OpenRouter terminal sequence: finish event, usage event with the same empty terminal choice, then `[DONE]`. Do not commit the raw capture, provider IDs, generated content, reasoning signatures, credentials, `.repro/`, or an absolute local path.
- Extend `PtcLlmHttpAdapterE2ETest` with an actual live **streaming** case. The current live test uses non-streaming `llm/request`, so it cannot detect #15. Assert ordered deltas, final content, normalized usage when supplied, successful completion, and absence of the credential from logs/inspection.
- Add a fresh-OS-process cold-hostname regression at the PtcRunner boundary. Resolver state must not already be warm. A hosts-file-backed hostname such as `localhost` is sufficient: the eventual public-policy address rejection is acceptable, but `resource_limit_exceeded` during DNS is not.
- Keep the adapter error mapping unchanged. `:malformed_stream` and `:resource_limit_exceeded` are being classified faithfully.
- Update the PR description and documentation so they distinguish deterministic coverage, live non-stream success, and live streaming success. Do not report the end-user path green until all three have run against the final exact dependency.
- Keep ReqLLM as the control during this PR; run the same manifest in the same session through both adapters and compare the stable result/usage shape.

These tests may be developed against local upstream fix branches, but no path/Git dependency or absolute checkout path should be committed.

### Join point after the upstream fixes

Once PtcLlmHttp publishes an exact patch containing #15 and #16:

1. Pin that exact version in `mix.exs`, update `mix.lock`, package verification, docs, and dependency tests.
2. Set PtcRunner's aggregate process budget to the documented upstream-supported cold-hostname value. If 40M remains necessary, document the per-attempt and maximum-concurrency memory exposure; otherwise remove the temporary 40M change.
3. Make the minimized terminal-usage replay and fresh-BEAM DNS regression green against the published package.
4. Run the documented `mix ptc run` command with `PtcLlmHttpAdapter` in a fresh OS process.
5. Run both live non-streaming and live streaming OpenRouter E2E tests with `OPENROUTER_API_KEY` and record the results in the PR.
6. Prefer the same patch release to include [ptc_llm_http#14](https://github.com/andreasronge/ptc_llm_http/issues/14), then replace direct opaque Error field reads with the public accessor and remove the Dialyzer suppression.

### Exit condition

The PR is ready to leave draft only when a fresh end-user command succeeds without resolver warming, a real OpenRouter stream completes after terminal usage, the deterministic replays cover both defects, no consumer workaround was added, and the exact published dependency passes package/release verification.

andreasronge:
Too old, will need to rework

Issue #1649, opened 2026-08-25
feat(llm): complete the public model contract and enforce operational budgets


andreasronge:
## Status

Implementation-ready tracking contract. A fresh independent design review and two
focused follow-ups approved the complete issue against `origin/main`
`df387aeaa5b22bcc93a4902fb70f3741ea1b8b1a` on 2026-08-25. No observable
contract questions remain. Implement as the eight ordered green child slices;
do not combine the parent into one PR. Each implementation branch still needs
the repository-required fresh cumulative code review.

### Implementation progress

- [x] Slice 1 — bounded dynamic-schema foundation: #1652
- [x] Slice 2 — requirement-aware preparation and per-call output authorization: #1656
- [x] Slice 3 — structured output
- [x] Slice 4 — ordinary whole-call deadline
- [x] Slice 5 — remaining inference controls and adapter parity
- [x] Slice 6 — usage guarantees and fixed-point observation
- [x] Slice 7 — aggregate total-token budget — optional-limit infrastructure in #1666; runtime ledger merged in #1679
- [x] Slice 8 — aggregate USD cost budget — tariff/preparation contract in #1667; runtime enforcement merged in #1679

### Remaining work

All eight slices are implemented. The runtime ledger for slices 7 and 8 landed in
PR #1679 (`aa0bf1c96`, 2026-08-28), with follow-ups #1699 (aggregate refusals as
runtime limits, `889f64eab`) and #1700 (optional budget diagnostics, `0b53c8a37`).
Verified on `main` `d98c4ba33`: `reservation_bound` (`llm.ex:105`), the opaque
`reference()` reservation id and `open_provider_gate/4` dispatch acknowledgement
(`run_state.ex:265`, `:307`), `finish_provider/3` settlement evidence
(`run_state.ex:320`), the authoritative `llm_budget` fact (`llm_budget.ex`,
projected at `command_run_outcome.ex:875`), and aggregate enforcement
(`limit_catalog.ex:125-128`, refused at `dispatcher.ex:1219`).

What closure still needs is the issue-level acceptance sweep, which is broader
than the ledger: replay exclusion, the applicable fault matrix, the runtime,
terminal, trace, command and Viewer projections authored from RunState,
generated artifacts, docs, and a fresh cumulative code review. Those were not
re-verified when the blocker paragraph above was replaced.

The parent remains open until all issue-level acceptance criteria are satisfied.

This tracks the **Model protocol** P1 in
[...]

PR #1657, opened 2026-08-26
feat(llm): add structured output envelopes

andreasronge:
Slice 3 of #1649. Do not treat this as the parent contract.

This PR adds live installation `structured_output_mode`, `llm/request` schema admission, and closed `structured_output` envelopes. Encoded structured `content` is deleted.

## What changed

- Live LLM installations require `structured_output_mode`: `"json_schema" | "json_object" | "unsupported"`. Changing it requires a new `installation_revision`. Replay has no mode.
- `llm/request` admits an optional top-level JSON object `schema` in the Kernel JSON Schema profile; it counts toward the argument-byte ceiling.
- Schema plus a non-empty `tools` list is `invalid_arguments` before reservation or dispatch.
- A proven-invalid request schema is `invalid_arguments` even under `unsupported`. A compiled valid schema under `unsupported` is `provider_error` / `structured_output_unsupported` before reservation.
- Adapter `json_schema` forces provider-native schema options and refuses tool-fallback-only targets: Groq, Bedrock, Azure, and OpenAI-compatible Vertex MaaS (GLM). Vertex Claude/Gemini, OpenRouter, OpenAI models with native schema, Anthropic, Google, Fireworks, and xAI remain eligible.
- Adapter `json_object` is attested separately: OpenAI-style `response_format` `json_object` on OpenRouter, OpenAI, Groq, Fireworks, xAI, Azure OpenAI, and Vertex OpenAI-compatible MaaS. Anthropic, Bedrock, Google AI Studio, Vertex Claude/Gemini, and Azure Claude are refused at prepare. Neither mode is a fallback for the other.
- Direct `ollama:` and `openai-compat:` HTTP routes are refused for both structured modes. The unreachable `:ollama` json_schema allow-list entry is gone.
- Kernel bounded-decodes `json_object` bytes, validates the object against the compiled request schema, and promotes a closed public envelope `{"structured_output": {}, "tokens": {}}` (plus router-injected `model`). There is no encoded `content` duplicate.
- Missing, malformed, oversized, non-object, or malformed-`tokens` structured candidates become `invalid_result` / `output_schema_mismatch`. That includes a well-formed object with invalid usage: usage guarantees are slice 6 (`promised_usage_missing`), so this slice keeps the closed-envelope mismatch. Decoder/validator heap/crash/deadline remains `output_validation_unavailable`.
- Replay fixtures match on schema and return the public `structured_output` envelope.
- Probe/doctor structured mode stays `:unsupported`. Public snapshots, discovery rows, and `models` installations omit `structured_output_mode`; the installation digest still hashes the decoded install map.
- Live descriptors seal `structured_output_mode` so acquisition reconstructs the same workflow route the prepare callback reports.
- Host examples, docs, prelude, generated host schema/site pages, and the release-smoke host are updated.

## Non-goals (later slices)

Deadlines, remaining inference controls, usage guarantees, and aggregate token/cost budgets stay on later #1649 slices. Usage/reservation remain interim-disabled.

The extra public `model` key on all three `llm/request` envelopes is pre-existing router injection and is not special-cased here.

## Verification

- `mix precommit`
- Focused structured-output, host-install, adapter (including wire `response_format`, no synthetic tools, json_schema Groq/Bedrock/Vertex GLM/Azure refusal, json_object Anthropic/Bedrock/Google/Vertex Gemini/Azure Claude refusal, direct ollama refusal of both modes), replay, digest, ManifestRepl worker-death, and CommandEngine live-LLM tests
- GitHub CI on this revision

<div><a href="https://cursor.com/agents/bc-b4737684-12f6-4d48-a929-e2a549172273?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b4737684-12f6-4d48-a929-e2a549172273&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


andreasronge:
## Independent review (codex, effort `medium`)

Reviewed `ca00e983` against `origin/main` `0a4d062b`. Base was unchanged before and after the run.

**Gate: FAIL — 3 critical findings.** All three verified against the source below.

---

### [P1] `json_object` is admitted for providers that cannot honor it

`json_object` is admitted for every ReqLLM-routed provider, but the implementation only supplies the OpenAI-style `response_format`. `req_llm_adapter.ex` `put_json_object_format/1` inserts `%{type: "json_object"}`, while `attest_structured_mode(model, :json_object)` accepts any `{:req_llm, _selector}` route.

Verified in the vendored dependency:

- `deps/req_llm/lib/req_llm/providers/anthropic.ex:158` lists `response_format` in `@unsupported_parameters`.
- `deps/req_llm/lib/req_llm/providers/amazon_bedrock.ex` contains no occurrence of `response_format` at all, and Bedrock's native request format has no equivalent field.

So an `anthropic:` or `amazon_bedrock:` alias declaring `structured_output_mode: "json_object"` either fails outright or dispatches ordinary text generation — after which the Kernel validates output the provider was never asked to shape. Every such call is billed and then rejected.

`json_object` needs the same per-provider/per-model attestation `json_schema` already has, using each provider's actual JSON-object control rather than one OpenAI-shaped option. Only providers that genuinely honor it should be admitted.

### [P1] Oversized structured responses get the wrong taxonomy

`llm_capability.ex` `normalize_response/3` classifies a response exceeding `max_response_bytes` as `ProviderError.new(:invalid_request, "LLM response exceeded its boundary")`, which `dispatcher.ex` `normalize_result/5` publishes as `provider_error` / `invalid_request`.

Section 11 of #1649 requires oversized provider output on the structured path to be:

```elixir
%{status: :error, kind: :invalid_result, reason: :output_schema_mismatch, retryable?: false}
```

The `:invalid_request` classification is inherited from the pre-slice-3 path, where it was correct. Slice 3 introduces the contract row that makes it wrong for schema requests, so it belongs in this slice.

### [P1] Malformed adapter success shapes escape as provider errors

The `{:ok, _wrong_branch}` arms in `dispatch_structured/4` and `call_req_llm_json/3` return `ProviderError.new(:invalid_result, ...)`. `normalize_result/5` maps every `%ProviderError{}` to `kind: :provider_error` with `reason: error.kind`, so these surface as `provider_error` / `invalid_result`.

The contract requires `kind: :invalid_result`, `reason: :output_schema_mismatch` — both fields differ. Section 12 is explicit that returning the wrong adapter branch is an invalid *result*, not a provider error.

This shares a root cause with the finding above: the structured path routes Kernel-side rejections through `ProviderError`, which cannot express the required `kind`. Either return a structured candidate for Kernel admission, or translate explicitly at the capability boundary.

---

### [P2] Structured success carries an extra `model` key — but this is pre-existing

Codex flagged that live structured success contains `model` alongside `structured_output` and `tokens`, contradicting section 12 ("Keys are closed... No `finish_reason`, output-limit record, encoded structured `content`, provider payload, or other metadata is public in V1"), and that `dispatcher_structured_output_test.exs` enshrines it.

Its suggested fix — suppress route result attributes for schema responses — would be wrong here. `lib/ptc_runner/kernel/llm_router.ex:151` sets `result_attributes: %{"model" => route.alias}` on `origin/main` today, and it applies to **all three** `llm/request` envelopes, not just the structured one. Suppressing it only for schema responses would make structured output inconsistent with the text and tool envelopes.

Real deviation from section 12, but not introduced by this slice. It should be resolved once for all three envelopes — either by reconciling section 12 with the router, or by removing the attribute everywhere — rather than special-cased here.

---

### Not found

No path was identified that reaches provider dispatch, credential access, or reservation ahead of a refusal the contract places before it. `structured_output_mode` was not observed leaking into public snapshots, discovery rows, or `models` output. The per-model `json_schema` allow-list (`ModelHelpers.json_schema?/1` for `:openai`, `XAI.supports_native_structured_outputs?/1` for `:xai`, family checks for `:google_vertex`) correctly handles the per-model capability differences within those providers.

---

Review prompt covered: pre-refusal dispatch/credential/reservation paths, failure-taxonomy misclassification in both directions, `json_object` provider viability, `structured_output_mode` projection leaks, and per-model vs per-provider allow-list correctness.

andreasronge:
## Independent review round 2 (codex, effort `medium`)

Fresh cold review of `5c3f64e6` against `origin/main` `0a4d062b`. Base unchanged before and after.

**Gate: FAIL — 1 critical finding.** Two of the three round-1 P1s are fully resolved.

---

### Round 1 status

**[P1] `json_object` admitted for providers that cannot honor it — FIXED.** The new `@native_json_object_providers` allow-list (`azure`, `fireworks_ai`, `groq`, `openai`, `openrouter`, `xai`) plus `attest_prepared_json_schema/2` for `:json_object` now refuses `anthropic`, `google`, `amazon_bedrock`, and `ollama` at prepare.

I checked the `google` refusal specifically, since Gemini does support JSON-object output. It is correct: `deps/req_llm/lib/req_llm/providers/google.ex:1218` sets `responseMimeType: "application/json"` only on the `:object` operation, which requires a `compiled_schema`. The `:response_format` occurrences at lines 459 and 564 are image-generation options. A `:chat` operation carrying `response_format` as a provider option would drop it, so refusing `google` for `json_object` is right.

**[P1] Oversized structured responses — FIXED.** `schema_output_mismatch/2` now returns `{:ok, %{}}` on the schema path so Dispatcher classifies it. I confirmed the empty map cannot surface as a successful empty result: `dispatcher.ex:1423` looks up the `"structured_output"` key and the `_missing` arm returns `{:error, :output_schema_mismatch}`.

**[P1] Malformed adapter success shapes — FIXED.** The `{:ok, _wrong_branch}` arms now pass the value up for Kernel admission instead of manufacturing a `ProviderError`.

---

### [P1] Malformed token usage on a structured response still gets the wrong taxonomy

`admit_tokens/1` in `llm_capability.ex` is reached from `normalize_response/3` at line 131, after `classify_success/2` has already succeeded. On malformed usage it returns:

```elixir
{:error, ProviderError.new(:invalid_request, "LLM response contained invalid token usage")}
```

unconditionally — it never consults `schema_request?`. `dispatcher.ex` `normalize_result/5` then publishes `kind: :provider_error`, `reason: :invalid_request`.

This is the same defect class as the two just fixed; the fix routed `normalize_response/3`'s own two branches through `schema_output_mismatch/2` but left this third exit untouched. A structured request whose provider returns a well-formed object with malformed `tokens` still escapes as a provider error.

`kind: :provider_error` is wrong under any reading of the contract. The exact target reason is worth a decision: codex proposes `invalid_result` / `output_schema_mismatch` (section 11's malformed-provider-output row), but section 11 also carries a distinct `invalid_result` / `promised_usage_missing` row, which is slice 6's. Since usage guarantees are interim-disabled in this slice, `output_schema_mismatch` looks right for now, but please make it a deliberate choice rather than an accident of which helper the branch calls.

Test coverage is missing for valid structured data with malformed `tokens`.

---

### [P2] Ollama is refused in both structured modes, and the allow-list entry is dead code

`parse_provider("ollama:" <> model_name)` returns `{:ollama, model_name}`, never `{:req_llm, _}`. Both `attest_structured_mode/2` branches only admit `{:req_llm, selector}` routes, so every `ollama:` selector is refused for both `json_schema` and `json_object`.

That makes the `:ollama` entry in `@native_json_object_providers`' sibling list `@native_json_schema_providers` unreachable — it can never be consulted. ReqLLM 1.20's Ollama provider does support native structured formatting, so the refusal is likely a false negative rather than a correct one.

`test "refuses json_schema preparation for ollama"` currently codifies the refusal rather than constraining intended behavior. Either implement the equivalent direct `format` request on the Ollama route, or drop `:ollama` from the allow-list and note the refusal as deliberate so the dead entry does not read as support.

### [P2] Structured success carries an extra `model` key — unchanged from round 1

Same finding, same position: `lib/ptc_runner/kernel/llm_router.ex:151` sets `result_attributes: %{"model" => route.alias}` on `origin/main` and applies it to all three `llm/request` envelopes. Suppressing it only for schema responses would make structured output inconsistent with the text and tool envelopes. Real deviation from section 12, but not this slice's to fix in isolation.

---

### Note, not a finding

`native_json_object_selector?/1` admits Vertex fail-open: `{:ok, :google_vertex, model_id} -> not vertex_native_json_schema_id?(model_id)`. Any Vertex model that is neither `claude-*` nor `gemini-*` is assumed to support `json_object`, including models that do not exist yet. That is defensible if every Vertex MaaS target is OpenAI-compatible by construction, but it is the one place on this path that defaults to admit rather than refuse. Worth an explicit comment either way.

---

Round-2 prompt covered the same five areas as round 1, plus whether the new tests constrain behavior or merely assert the current implementation.

PR #1661, opened 2026-08-26
feat(llm): enforce ordinary whole-call request deadlines

andreasronge:
Slice 4 of #1649. Do not treat this as the parent contract.

This PR adds the identity-bearing `llm_request_timeout_ms` limit, live installation `ceilings.request_timeout_ms`, and one Kernel-enforced absolute whole-call deadline for ordinary live model calls.

## What changed

- LimitCatalog row `llm_request_timeout_ms`: compiled and installed default `120_000`, maintained `100..1_800_000`, `manifest_narrowable`. Hosts may widen or narrow the installed ceiling; applications may only narrow.
- Live LLM installations may set `ceilings.request_timeout_ms` in `100..host installed ceiling`. Omission means `120_000` clamped by that ceiling. Changing it requires a new `installation_revision`. Replay has no ceiling.
- Dispatcher samples `now` once immediately before `RunState.reserve_capability` and stores a private combined `llm_request_deadline_ms` as `min(LLM clocks, enclosing run/workflow clocks)`.
- Live `source: "llm"` requesters receive that integer deadline in the closed arity-two context `%{llm_request_deadline_ms: integer() | nil}`. Replay and the doctor connectivity probe receive `nil`.
- The Dispatcher-owned provider worker kills hung adapters at the combined cutoff. The built-in adapter derives remaining time immediately before ReqLLM, passes `total_timeout`, and keeps `receive_timeout` no later. Retries share the same absolute deadline.
- When the LLM clock wins, the public result is retryable `timeout/llm_request_timeout`. Adapter-returned timeouts after that clock wins are reclassified the same way. Equal enclosing clocks keep existing taxonomy (`run_deadline` / `provider_timeout` / `output_validation_unavailable`). Equal LLM and shared-validation clocks are LLM timeout. Schema compile stays on the shared/enclosing clocks.
- Generated host/manifest/command schemas, kernel-limits docs, host-installation docs, and site pages include the new row.

## Non-goals (later slices)

Remaining inference controls, usage guarantees, and aggregate token/cost budgets stay on later #1649 slices.

## Verification

- `mix precommit`
- `scripts/ci/core-tests.sh` — 7212 passed
- `scripts/ci/core-dialyzer.sh`
- Focused dispatcher deadline, catalog, host-config, host-installation, adapter elapsed-deadline, structured-output, router, dispatcher-effect, and replay requester tests
- Independent reviews (3, max 4): incremental approve, cumulative base-guarded approve, follow-up on the replay arity-two test callers approve
- GitHub CI on this revision

<div><a href="https://cursor.com/agents/bc-01a03f61-4e91-7139-867d-3ad44349c652?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a03f61-4e91-7139-867d-3ad44349c652&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


lib/ptc_runner/kernel/dispatcher.ex
```diff
@@ -364,6 +370,98 @@ defmodule PtcRunner.Kernel.Dispatcher do
...
+
+  defp enclosing_remaining_ms(%CapabilityInvocation{enclosing_deadline_ms: deadline})
+       when is_integer(deadline),
+       do: max(deadline - System.monotonic_time(:millisecond), 0)
+
+  defp enclosing_remaining_ms(_invocation), do: 1_000_000_000
+
+  defp llm_deadline_wins?(%CapabilityInvocation{llm_request_deadline_ms: deadline} = invocation)
+       when is_integer(deadline),
+       do: llm_remaining_ms(invocation) == 0 and enclosing_remaining_ms(invocation) > 0
```
andreasronge: [P2] Preserve the deadline winner from the absolute clocks

This determines the winner from which clocks are still live at observation time. If the LLM deadline is earlier than the enclosing deadline but the Dispatcher is not scheduled until after both have expired, both remaining values are zero and the call is reported as `provider_timeout`, even though the LLM clock objectively won. That is realistic under load or with close deadlines and makes the taxonomy scheduler-dependent. Check that the LLM deadline is expired and compare the stored absolute deadlines (`llm_deadline < enclosing_deadline`, with equality retaining the enclosing taxonomy); output validation needs the analogous absolute comparison with its shared deadline.

lib/ptc_runner/kernel/dispatcher.ex
```diff
@@ -364,6 +370,98 @@ defmodule PtcRunner.Kernel.Dispatcher do
...
+  defp await_timeout_result(invocation) do
+    if llm_deadline_wins?(invocation) do
+      llm_request_timeout()
+    else
+      %{status: :error, kind: :timeout, reason: :provider_timeout, retryable?: true}
+    end
+  end
+
+  defp llm_request_timeout do
+    %{status: :error, kind: :timeout, reason: :llm_request_timeout, retryable?: true}
```
andreasronge, line 448: [P1] Teach agent.core about this new timeout envelope

The shipped `agent.core/run-outcome` classifier only treats `kind == "timeout"` as recoverable when the reason is `provider-timeout`/`provider_timeout` (`priv/preludes/kernel/agent.core.clj:321-322`). This new `timeout/llm_request_timeout` therefore takes the fail-fast branch at lines 699-706 instead of returning the documented `:provider-failure`, despite `retryable?` being true. Any live call that reaches this new limit breaks the caller failover/retry contract. Please add the new reason to the prelude classifier and cover it in `agent_library_test.exs` (and regenerate the prelude docs/artifacts).

lib/ptc_runner/llm/req_llm_adapter.ex
```diff
@@ -1359,9 +1370,42 @@ if Code.ensure_loaded?(ReqLLM) do
...
+      |> Keyword.put(:total_timeout, remaining)
+      |> Keyword.put(:receive_timeout, receive_timeout)
+    end
+
+    defp req_timeout_opts(opts) do
+      receive_timeout = Keyword.get(opts, :receive_timeout, @default_timeout)
+
+      case Keyword.get(opts, :total_timeout) do
+        total when is_integer(total) and total > 0 ->
+          [receive_timeout: min(receive_timeout, total), total_timeout: total]
```
andreasronge, line 1426: [P2] Normalize ReqLLM total-timeout results as timeouts

Passing `total_timeout` makes ReqLLM return `%ReqLLM.Error.API.Timeout{}` when that budget expires, but `normalize_provider_error/1` has no clause for this struct, so it falls through to `ProviderError.new(:unavailable, ...)`. I reproduced that on this head with `ReqLLMAdapter.normalize_provider_call({:error, ReqLLM.Error.API.Timeout.exception(kind: :total, timeout: 10)})`. If that result reaches the Dispatcher before its `after` branch is selected, `provider_error_result/3` cannot reclassify it because its kind is `:unavailable`, and the public result becomes `provider_error/unavailable` instead of `timeout/llm_request_timeout`. Please map ReqLLM total-timeout exceptions to a retryable `ProviderError` timeout and add a focused test.

Issue #1642, opened 2026-08-25
research(analysis): admit bounded private cohorts without loading every payload


andreasronge:
## Context

Follow-up to #1636. Exact named-run selection can open one canonical trace/inspection pair without scanning or loading unrelated private history. Private cohort work still needs a bounded way to discover or admit several runs without compiling every private payload in an accumulating artifact root.

This is distinct from #1639:

- #1639 concerns model-attributed public cohort summaries already derivable from canonical traces.
- This issue concerns private source admission and metadata discovery before opening selected private payloads.

## Problem

`private-run-analysis-v1` currently captures whole trace and inspection directories before `analysis/runs` can page or filter them. Output pagination cannot bound that source-wide inventory, parsing, correlation, and retention work.

A cohort mechanism must support many available runs while keeping each private payload closed until explicitly selected. It must remain immutable, replayable, fail-closed, and bounded without becoming a mutable authoritative run database.

### Immutable per-run metadata

Derive bounded non-payload metadata from sealed per-run trace/inspection manifests:

- run and trace identities;
- artifact presence and source class;
- schema/version compatibility;
- byte and record counts;
- safe canonical status and timestamps;
- evidence digest and correlation state;
- no prompts, responses, generated source, capability payloads, diagnostics, prints, or result values.

### Source-bound cohort catalog

A catalog may be an immutable generation, content-addressed manifest tree, or a bounded selection supplied by the operator. It must identify exactly which per-run manifests belong to the captured cohort and commit to their identities.

Pagination bounds returned metadata but is not itself an immutable filesystem snapshot. Cursor validity therefore binds a catalog generation or digest, not a live root listing.

### Open payloads separately

Metadata discovery and private payload opening are distinct authorities:

1. list/filter safe cohort metadata;
2. select one run or a bounded run-ID set;
3. open only those correlated private payloads through the selected snapshot contract from #1636;
4. query each payload under existing private-output and result controls.

A PTC-Lisp session may process a cohort incrementally, but every operation retains finite call, result, heap, disk, handle, and time limits.

## Research questions

1. What metadata is safe and sufficient to select private cohorts without opening private payloads?
2. Is the catalog an immutable generation, content-addressed tree, explicit run-ID set, or more than one supported source form?
3. How is a new run published without mutating the identity of an already-open catalog?
4. How are trace-only, inspection-only, incompatible-schema, and failed-correlation runs represented safely?
[...]

PR #1680, opened 2026-08-28
feat(analysis): capture bounded private run-catalog generations

andreasronge:
Refs #1642 — implements **step 1 of the four-PR breakdown** frozen in [the design-decisions comment](https://github.com/andreasronge/ptc_runner/issues/1642#issuecomment-5445118504): the `RunCatalog` capture owner. Steps 2–4 remain; see **Remaining work** below.

> Current head `5d6921e`. The review of the first revision raised six findings, all fixed; every review thread is resolved and no question is outstanding. See **Review history** below.

## The frozen contract this implements

Per §1, §2, §4 and §5 of the design comment, as amended by [the accepted schema amendment](https://github.com/andreasronge/ptc_runner/issues/1642#issuecomment-5449287788):

- **Derived immutable snapshot, no sidecars, no persisted catalog file.** A generation is an in-memory capture owned by `RunCatalogSnapshot`, built from bounded probes: a `.ptcins` container's 16-byte header and 192-byte footer (two `pread`s, no frame decoded), and a canonical trace's first line plus the last complete line inside a bounded tail window.
- **Identity is verified, never trusted from a filename.** `sha256(stem) == footer.run_id_sha256` and `sha256(trace head trace_id) == footer.trace_id_sha256`; a disagreement isolates the row. The two probed trace events are additionally validated by `TraceEventValidation`, so a file the canonical reader rejects is never published as admissible.
- **The bytes read are the bytes committed to.** Each file is `lstat`ed before the open, the opened descriptor is `fstat`ed and compared before a byte is read, and the path is `lstat`ed again after — so neither a file replaced under its path nor one rewritten mid-probe is described as the file that was listed.
- **Generation digest.** `sha256("ptc-run-catalog-v1\0" <> sorted per-row commitments)`, where each commitment is `{run_ref, trace half, inspection half}` and each half is `{present, descriptor identity, sha256 of the exact bytes probed}`. Committing to *bytes* rather than to decoded values is load-bearing: a footer's stored `artifact_digest` is a claim a metadata-only probe never recomputes, so a commitment derived from it misses a rewrite of any other footer field. Halves that read bytes but did not decode — malformed, foreign-version — commit to them too; only a half that never reached any bytes carries an absent marker. Nothing re-lists after capture, so a growing root cannot alter an open generation.
- **Row schema (§2)** as specified, including the exclusions: no prompts, responses, generated source, capability payloads, diagnostics, prints, result values, evidence content, filesystem paths, or any field needing a whole-trace scan. Plus the two amended enum values, `correlation: "unavailable"` and `trace_present: "unreadable"`, for an entry whose probe failed on a half.
- **Failure matrix (§4).** Per-row isolation with path-free codes — `ambiguous_trace`, `malformed_metadata`, `unstable_entry`, `unsupported_schema`, `filename_run_mismatch`, `duplicate_run_identity` — and whole-operation refusals `source_unavailable` / `catalog_limit_exceeded`. A malformed *unselected* entry never blocks the generation. Every duplicate claimant is isolated; the primary reason follows existing precedence.
- **Limits (§5).** 4,096 directory entries, 1,024 stems, 65,536-byte head + 65,536-byte tail + 208 sealed edge bytes per file, 2,048-byte rows, 4,194,304 retained bytes, 15,000 ms capture deadline, 5,000 ms listing deadline. All may be lowered by a caller and none raised.

## Decisions the frozen comment left unstated

1. **`correlation: "unavailable"` and `trace_present: "unreadable"`** — values beyond the original enums, for an entry whose probe failed on a half. **Proposed on #1642 and accepted into the frozen contract.**
2. **Correlation-mismatch row reason** — reuses the existing `inspection_correlation_missing`, the code the matrix already assigns to that state on the selection side.
3. **Oversized-row reason** — the reduced row is isolated as `malformed_metadata`.
4. **Header/footer version disagreement is `malformed_metadata`, not `unsupported_schema`** — an internally inconsistent container is corruption, not a version this build cannot read.
5. **A tail `run-stopped` naming a different run is `malformed_metadata`** — the two lines disagree about whose file it is.
6. **A non-integer `schema_version` is `malformed_metadata`, not `unsupported_schema`** — it is not a version this build cannot read, it is an event the canonical reader rejects.
7. **Identity strings bounded to 256 bytes** (matching `TraceDirectoryAdmission`'s existing claim bound) **and timestamps to 64 bytes.** ISO 8601 admits an unbounded fractional-second field, so a parse alone is not a bound; the row byte bound cannot drop either field, so both are bounded at the probe.
8. **Extracted `BoundedCapture`.** The duplication gate flagged the owner-capture loop as a copy of `TraceSnapshot`'s; both now share one owner-scoped implementation, each mapping its outcome onto its own source codes.

## Verification

Toolchain had to be installed in this container (`mise install` → Erlang 29.0.3 / Elixir 1.20.2, per `mise.toml`).

| Command | Result on `5d6921e` |
|---|---|
| `mix compile --warnings-as-errors` | clean |
| `mix precommit` | passes; duplication `new=0` |
| `mix dialyzer --format short` | `Total errors: 0` |
| `MIX_ENV=dev mix docs --warnings-as-errors` | clean (it caught a moduledoc link to a hidden module) |
| `mix test` (catalog files) | 45 passed |
| GitHub CI | all checks green |

The full suite in this container reports seven **pre-existing environment failures** unrelated to this diff: re-running those files on clean `main` in the same container reproduces the identical seven. Four are `chmod`-based tests asserting `:eacces`, defeated by the container running as root; two need IPv6 (`:eafnosupport`); one REPL frontend test is in the same class. GitHub CI passes all of them, which confirms the diagnosis.

The repo's pre-push hook is not installed in this fresh clone (`core.hooksPath` unset), so each gate it invokes was run by hand.

## Review history

Six reproduced findings on the first revision, all fixed. Each reproduction is now a test.

**P1 — probed events were not validated as canonical trace events.** Projecting from a handful of hand-picked fields let a file the canonical reader rejects be published as complete and admissible. The two probed events now go through `TraceEventValidation`. `{:error, :unsupported_version}` still projects so the row can report its declared version; every other failure isolates the row.

**P1 — the generation digest was bound to decoded values.** Rewriting `record_count` in a footer changed the row while leaving `catalog_digest` identical, because the commitment was built from the footer's stored `artifact_digest`. Both halves now commit to the exact bytes probed plus descriptor identity.

**P2 — the opened descriptor was never `fstat`ed.** A path replaced between `lstat` and `open` and restored before the closing `lstat` could commit replacement bytes under the original identity. Fixed following `Handle.open/2`; this required dropping `:raw`, since a raw descriptor is process-local and cannot be `fstat`ed. I judged the window untestable without a `Process.sleep` or a test-only export; the maintainer showed otherwise in `5d6921e`, adding a nil-by-default `:file_probe_hook` on the same footing as the existing `:listing_hook` and a rendezvous-based regression that swaps the file before `open` and restores it before the closing `lstat`, so only the `fstat` can catch it.

**P2 — current-version header geometry was unvalidated.** A declared header size of 17 was accepted although the artifact reader rejects it. `Format.decode_header/1` now validates the whole header once the versions are recognised as this build's.

**P2 — duplicate detection ignored inspection-footer identity claims.** Traces state identities in plaintext and footers only as digests, so the two halves claimed into spaces that could never intersect, and copying one run's artifact over another's canonical name left the impersonated run admissible. The plaintext is now hashed to meet the footer.

**P2 — classifier inputs were only partially validated.** The partial check advertised `{:error, :invalid_catalog}` and then broke it by raising out of a projection. Every field the classifier reads is now validated.

## Remaining work

Steps 2–4 of the frozen breakdown are untouched by this PR:

2. **`analysis-catalog` capability + `private-run-catalog-v1` profile** — paging, the filter vocabulary plus `correlation`/`state`, digest-bound cursors reusing the `TraceLog` cursor helpers, `excluded_files` accounting, REPL/JSONL wiring. `RunCatalogSnapshot.rows/1` is the frozen list that surface will page over.
3. **Bounded multi-run selected capture** — `{:selected_canonical_set, dir, run_refs}` in `SelectedCanonicalSource` and both snapshots, repeated `--run` on `private-run-analysis-v1`, `max_selected_runs` (16), aggregate-limit enforcement, and the `duplicate_selected_run` / `selected_set_limit_exceeded` / isolated-run refusals.
4. **Docs, diagnostics, integration** — `docs/reference/repl.md`, `cli.md`, `docs/maintainers/trace-log-contract.md`, `debug-navigation.md`, stable code-catalog additions.

Because the catalog has no user-facing surface until step 2, this PR adds no user documentation; the contract lives in the three module docs. Acceptance-list coverage is correspondingly partial: catalog growth during paging, stale cursors, and bounded incremental opening are step 2/3 concerns. Covered here: duplicate identities across both halves, missing pairs, changed manifests, malformed entries selected and unselected, non-canonical events, descriptor replacement under the inventoried path, large catalogs, generation immutability under a growing root, digest stability and change, payload-content leak checks, and owner cleanup.

### DIFF
--- a/lib/ptc_runner/kernel/run_catalog.ex
+++ b/lib/ptc_runner/kernel/run_catalog.ex
@@ -109,8 +109,16 @@ defmodule PtcRunner.Kernel.RunCatalog do
   end
 
   # Two entries may not both claim one run or trace identity. Stems are unique
-  # within a root, so this catches the case the filesystem cannot prevent: a
-  # trace whose embedded identity belongs to another entry's run.
+  # within a root, so this catches the cases the filesystem cannot prevent: a
+  # trace whose embedded identity belongs to another entry's run, and a sealed
+  # artifact copied under a second run's canonical name.
+  #
+  # Both halves claim into one space. A trace states its identities in
+  # plaintext and a footer states them only as digests, so the plaintext is
+  # hashed to meet the footer rather than the footer being trusted to name
+  # itself — otherwise the two halves would claim into spaces that never
+  # intersect, and an artifact duplicating another run's identity would leave
+  # the run it impersonates looking admissible.
   defp duplicated_identities(probes) do
     probes
     |> Enum.flat_map(&claimed_identities/1)
@@ -120,10 +128,23 @@ defmodule PtcRunner.Kernel.RunCatalog do
     |> MapSet.new()
   end
 
-  defp claimed_identities(%{run_ref: run_ref, trace: %{present: :probed, head: head}}),
-    do: [{{:run, head.run_id}, run_ref}, {{:trace, head.trace_id}, run_ref}]
+  defp claimed_identities(%{run_ref: run_ref, trace: trace, inspection: inspection}) do
+    Enum.map(trace_claims(trace) ++ inspection_claims(inspection), &{&1, run_ref})
+  end
 
-  defp claimed_identities(_probe), do: []
+  defp trace_claims(%{present: :probed, head: head}) do
+    [
+      {:run, Format.identity_sha256(head.run_id)},
+      {:trace, Format.identity_sha256(head.trace_id)}
+    ]
+  end
+
+  defp trace_claims(_trace), do: []
+
+  defp inspection_claims(%{present: :probed} = inspection),
+    do: [{:run, inspection.run_id_sha256}, {:trace, inspection.trace_id_sha256}]
+
+  defp inspection_claims(_inspection), do: []
 
   defp row(probe, duplicated) do
     correlation = correlation(probe)
@@ -313,26 +334,20 @@ defmodule PtcRunner.Kernel.RunCatalog do
     end
   end
 
-  # A row commitment covers the entry's reference, the trace's filesystem
-  # identity, the digest of the bytes actually probed, and the sealed artifact
-  # digest. Absent halves commit to their absence, so a generation captured
+  # A row commitment covers the entry's reference and, for each half, its
+  # classification, the descriptor identity it was read from, and a digest of
+  # the exact bytes probed. Committing to decoded values instead would miss a
+  # rewrite the decode carries through unchanged — a footer's stored
+  # `artifact_digest` is a claim a metadata-only probe never recomputes, so
+  # editing a footer field beside it changes the row while leaving the claim
+  # alone. Absent halves commit to their absence, so a generation captured
   # before a run was published cannot collide with one captured after.
   defp commitment(%{run_ref: run_ref, trace: trace, inspection: inspection}) do
-    {run_ref, trace_commitment(trace), inspection_commitment(inspection)}
+    {run_ref, half_commitment(trace), half_commitment(inspection)}
   end
 
-  defp trace_commitment(%{present: :probed, identity: identity, commitment: commitment}),
-    do: {:probed, identity, commitment}
-
-  defp trace_commitment(%{present: present}), do: {present}
-
-  defp inspection_commitment(%{present: :probed} = inspection),
-    do: {:probed, inspection.bytes, inspection.artifact_digest}
-
-  defp inspection_commitment(%{present: :versions} = inspection),
-    do: {:versions, inspection.bytes, inspection.format_version, inspection.schema_version}
-
-  defp inspection_commitment(%{present: present}), do: {present}
+  defp half_commitment(%{present: present} = half),
+    do: {present, Map.get(half, :identity), Map.get(half, :commitment)}
 
   defp digest(commitments) do
     [@catalog_version, 0, :erlang.term_to_binary(Enum.sort(commitments), [:deterministic])]
@@ -351,11 +366,15 @@ defmodule PtcRunner.Kernel.RunCatalog do
 
   defp valid_probe?(_probe), do: false
 
+  # Every field the classifier goes on to read is checked here, not a sample of
+  # them. A partial check is worse than none: it reports a contract
+  # (`{:error, :invalid_catalog}`) that it then breaks, raising out of a
+  # projection instead — an untyped `head.timestamp` reaching
+  # `DateTime.from_iso8601/1` is the shape that takes.
   defp valid_trace_probe?(%{present: :probed, head: head, tail: tail} = trace) do
-    Map.has_key?(trace, :bytes) and Map.has_key?(trace, :identity) and
-      Map.has_key?(trace, :commitment) and trace.source_kind in [:sanitized, :private] and
-      is_map(head) and is_map(head.labels) and is_binary(head.run_id) and
-      is_binary(head.trace_id) and (is_nil(tail) or is_map(tail))
+    non_neg_integer?(trace[:bytes]) and is_tuple(trace[:identity]) and
+      is_binary(trace[:commitment]) and trace[:source_kind] in [:sanitized, :private] and
+      valid_head?(head) and valid_tail?(tail)
   end
 
   defp valid_trace_probe?(%{present: present}),
@@ -363,21 +382,45 @@ defmodule PtcRunner.Kernel.RunCatalog do
 
   defp valid_trace_probe?(_trace), do: false
 
+  defp valid_head?(head) when is_map(head) do
+    is_binary(head[:run_id]) and head[:run_id] != "" and is_binary(head[:trace_id]) and
+      head[:trace_id] != "" and is_map(head[:labels]) and optional_binary?(head[:timestamp]) and
+      (is_nil(head[:schema_version]) or is_integer(head[:schema_version]))
+  end
+
+  defp valid_head?(_head), do: false
+
+  defp valid_tail?(nil), do: true
+
+  defp valid_tail?(tail) when is_map(tail) do
+    optional_binary?(tail[:timestamp]) and optional_binary?(tail[:status]) and
+      optional_binary?(tail[:terminal_reason]) and optional_binary?(tail[:result_hash])
+  end
+
+  defp valid_tail?(_tail), do: false
+
   defp valid_inspection_probe?(%{present: :probed} = inspection) do
-    Map.has_key?(inspection, :bytes) and Map.has_key?(inspection, :record_count) and
-      is_binary(inspection.run_id_sha256) and is_binary(inspection.trace_id_sha256) and
-      is_binary(inspection.artifact_digest)
+    non_neg_integer?(inspection[:bytes]) and non_neg_integer?(inspection[:record_count]) and
+      is_tuple(inspection[:identity]) and is_binary(inspection[:commitment]) and
+      is_integer(inspection[:format_version]) and is_integer(inspection[:schema_version]) and
+      is_binary(inspection[:run_id_sha256]) and is_binary(inspection[:trace_id_sha256]) and
+      is_binary(inspection[:artifact_digest])
   end
 
   defp valid_inspection_probe?(%{present: :versions} = inspection) do
-    Map.has_key?(inspection, :bytes) and is_integer(inspection.format_version) and
-      is_integer(inspection.schema_version)
+    non_neg_integer?(inspection[:bytes]) and is_tuple(inspection[:identity]) and
+      is_binary(inspection[:commitment]) and is_integer(inspection[:format_version]) and
+      is_integer(inspection[:schema_version])
   end
 
   defp valid_inspection_probe?(%{present: present}),
     do: present in [:absent, :malformed, :unstable]
 
   defp valid_inspection_probe?(_inspection), do: false
+
+  defp non_neg_integer?(value), do: is_integer(value) and value >= 0
+
+  defp optional_binary?(value), do: is_nil(value) or is_binary(value)
 
   defp unique_refs?(probes) do
     refs = Enum.map(probes, & &1.run_ref)
--- a/lib/ptc_runner/kernel/run_catalog.ex
+++ b/lib/ptc_runner/kernel/run_catalog.ex
@@ -1,0 +1,392 @@
+defmodule PtcRunner.Kernel.RunCatalog do
+  @moduledoc """
+  Immutable safe-metadata rows for one private run-cohort generation.
+
+  A generation is a captured value, not a view of a directory. It turns the
+  bounded probes of `PtcRunner.Kernel.RunCatalogProbe` into one frozen row per
+  canonical run stem, commits to their identities with a domain-separated
+  digest, and never re-lists anything afterwards. Filesystem change therefore
+  cannot alter an open generation: a later capture is a different generation
+  with a different digest.
+
+  ## What a row may say
+
+  A row carries only what a sealed artifact states about itself at O(1) cost:
+  identities, artifact presence and source class, schema and format versions,
+  byte and record counts, lifecycle timestamps and terminal status, the safe
+  `result_hash` fingerprint, the same `run-started` labels the public
+  `analysis/runs` surface already exposes, the sealed artifact digest, and the
+  correlation and admissibility of the pair.
+
+  A row never carries prompts, responses, generated source, capability
+  payloads, diagnostics, prints, result values, evidence-record content, or a
+  filesystem path — and never a counter that would require scanning a whole
+  trace (`evaluations`, `llm_calls`, `error_count`, `truncated`). Those stay
+  where they already are, behind admission, on `analysis/runs {"view" "full"}`
+  and `analysis/counters`.
+
+  "Non-copying" here means no payload copying: beyond a head line, a bounded
+  tail line, and a sealed container's fixed edges, nothing is read, decoded,
+  or retained. The bounded per-row metadata projection is the product.
+
+  ## Isolation
+
+  Classification is per row, and a row is the unit of failure. A malformed,
+  ambiguous, duplicate, unstable, or version-incompatible entry becomes an
+  isolated row carrying a path-free reason code — it never blocks the
+  generation, because nothing beyond its own probe was ever opened. Whole-
+  operation refusals belong to the probe and the owner, not here.
+  """
+
+  alias PtcRunner.Kernel.InspectionArtifact.Format
+  alias PtcRunner.Kernel.RunCatalogProbe
+
+  @catalog_version "ptc-run-catalog-v1"
+  @trace_schema_version 2
+  @max_row_bytes 2_048
+
+  # Most fundamental first: a reason that describes the entry itself outranks
+  # one that describes its relationship to another entry, so a row that is both
+  # malformed and duplicated reports the malformation it actually has.
+  @reason_order [
+    :ambiguous_trace,
+    :unstable_entry,
+    :malformed_metadata,
+    :unsupported_schema,
+    :filename_run_mismatch,
+    :duplicate_run_identity,
+    :inspection_correlation_missing
+  ]
+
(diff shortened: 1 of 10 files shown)


lib/ptc_runner/kernel/run_catalog.ex
```diff
@@ -0,0 +1,392 @@
...
+  defp commitment(%{run_ref: run_ref, trace: trace, inspection: inspection}) do
+    {run_ref, trace_commitment(trace), inspection_commitment(inspection)}
+  end
+
+  defp trace_commitment(%{present: :probed, identity: identity, commitment: commitment}),
+    do: {:probed, identity, commitment}
+
+  defp trace_commitment(%{present: present}), do: {present}
+
+  defp inspection_commitment(%{present: :probed} = inspection),
```
andreasronge: **[P1] Bind the generation digest to the exact footer bytes.** The row reads record count, identity hashes, and other metadata from the footer, but this commitment includes only file size and the footer's claimed artifact digest. That digest is not verified during metadata-only probing. I reproduced changing `inspection_record_count` from 11 to 12 while `catalog_digest` remained identical. Future cursors could therefore accept a different catalog result as the same generation. Commit to a digest of the exact opened header/footer bytes and descriptor identity, including malformed and foreign-version probes.

lib/ptc_runner/kernel/run_catalog_probe.ex
```diff
@@ -0,0 +1,511 @@
...
+    case candidate |> :binary.split("\n", [:global]) |> Enum.reverse() do
+      [_trailing | terminated] -> Enum.find(terminated, &(&1 != "")) |> wrap_line()
+      [] -> :error
+    end
+  end
+
+  defp wrap_line(nil), do: :error
+  defp wrap_line(line), do: {:ok, line}
+
+  defp decode_started(line) do
```
andreasronge: **[P1] Validate the probed events as canonical trace events before projecting them.** The head/tail parser validates only a few selected fields. After deleting `trace_id` from `run-stopped`, current-main `TraceEventValidation` returned `malformed_source`, while this catalog reported the run as complete and admissible. Missing schema, sequence, timestamp/data structure, current terminal usage fields, mismatched trace identity, and invalid result hashes have similar paths. Please factor out bounded canonical validation for the probed events.

lib/ptc_runner/kernel/run_catalog_probe.ex
```diff
@@ -0,0 +1,511 @@
...
+          _moved ->
+            %{present: :unstable}
+        end
+
+      :error ->
+        %{present: :malformed}
+    end
+  end
+
+  defp read_edges(path, size) do
```
andreasronge: **[P2] Verify that the opened descriptor is the inventoried file.** The probe lstats the path, opens it, reads it, and lstats the path again, but never fstats the opened descriptor. A path can be replaced between lstat and open and restored before the final lstat, causing replacement bytes to be committed with the original file identity. Fstat the opened descriptor and compare it to the initial stat, following `InspectionArtifact.Handle.open/2`.

lib/ptc_runner/kernel/run_catalog_probe.ex
```diff
@@ -0,0 +1,511 @@
...
+    header = binary_part(head, 0, Format.header_size())
+
+    footer_bytes =
+      binary_part(
+        tail_bytes,
+        byte_size(tail_bytes) - Format.footer_size(),
+        Format.footer_size()
+      )
+
+    with {:ok, header_versions} <- Format.decode_header_versions(header),
```
andreasronge, line 586: **[P2] Reject invalid current-version header geometry.** `decode_header_versions/1` ignores the declared header size, and the current-version path never calls `Format.decode_header/1`. I changed the header-size field from 16 to 17 and the catalog still returned a paired, admissible row, although the normal artifact reader rejects it as malformed. Once versions are recognized as current, validate the complete current header before accepting the footer projection.

andreasronge (commented):
Reviewed against updated `origin/main` at `1ad966ea`, not only the PR's original base. The synthetic merge was clean; the 38 new catalog tests and 50 impacted trace/snapshot tests passed. The inline comments below capture six reproduced or directly reachable correctness issues.

Contract drift to settle before step 2: the frozen #1642 schema lists four correlation values and three trace-presence values, while this PR adds `"unavailable"` and `"unreadable"`. Those may be sensible, but they should be added to the issue's frozen contract rather than living only in the PR body.

GitHub CI is green, but it does not cover the reproduced cases in the inline findings.

lib/ptc_runner/kernel/run_catalog.ex
```diff
@@ -0,0 +1,392 @@
...
+  defp duplicated_identities(probes) do
+    probes
+    |> Enum.flat_map(&claimed_identities/1)
+    |> Enum.group_by(&elem(&1, 0), &elem(&1, 1))
+    |> Enum.filter(fn {_identity, refs} -> refs |> Enum.uniq() |> length() > 1 end)
+    |> Enum.flat_map(fn {_identity, refs} -> refs end)
+    |> MapSet.new()
+  end
+
+  defp claimed_identities(%{run_ref: run_ref, trace: %{present: :probed, head: head}}),
```
andreasronge: **[P2] Include inspection-footer identity claims in duplicate detection.** This graph considers only plaintext identities from trace heads. It ignores `run_id_sha256` and `trace_id_sha256` claimed by inspection footers, despite the frozen failure matrix requiring all duplicate claimants to be isolated. Copying run A's `.ptcins` over run B's canonical inspection name left A admissible and isolated only B as a filename mismatch. Add the hashed inspection claims so every claimant is isolated.

lib/ptc_runner/kernel/run_catalog.ex
```diff
@@ -0,0 +1,392 @@
...
+  # honest answer to a probe the classifier cannot read: half a row would
+  # publish a state nothing observed.
+  defp valid_probe?(%{run_ref: run_ref, trace: trace, inspection: inspection}) do
+    is_binary(run_ref) and run_ref != "" and valid_trace_probe?(trace) and
+      valid_inspection_probe?(inspection)
+  end
+
+  defp valid_probe?(_probe), do: false
+
+  defp valid_trace_probe?(%{present: :probed, head: head, tail: tail} = trace) do
```
andreasronge, line 374: **[P2] Finish validating classifier inputs before building rows.** `valid_trace_probe?/1` checks only a subset of fields subsequently read by the classifier. A probe with `head.timestamp: 42` and a non-nil tail passes validation and then raises inside `DateTime.from_iso8601/1`, contradicting `generation/2`'s `{:error, :invalid_catalog}` contract. Validate the byte/count/commitment, head timestamp/schema, and complete tail field types here.

lib/ptc_runner/kernel/run_catalog_probe.ex
andreasronge: Fixed in 484ebaf. Rather than write a parallel validator I hand the two probed events to `TraceEventValidation.validate/1` — it already owns per-event shape, run/trace identity agreement, sequence ordering and run lifecycle, and validating exactly the two events the row is built from keeps it bounded. The projection now happens only after it returns.

One deliberate exception: `{:error, :unsupported_version}` still projects, so the row can report the version it declares rather than being called corrupt. Every other failure isolates the row with nothing projected.

Two consequences worth flagging. A `schema_version` that is not an integer is now `malformed_metadata` rather than `unsupported_schema` — it is not a version this build cannot read, it is an event the canonical reader rejects — and I updated that test accordingly. And the tail is only validated alongside the head when it is a `run-stopped`; for a run still being written I validate the head alone, because an arbitrary mid-run last line would drag `advance_evaluation_lifecycle` into judging events the probe never read.

Your reproduction is now a test: deleting `trace_id` from `run-stopped` isolates the row. I added a second for the result-hash-contradicting-outcome path you mentioned.

---

lib/ptc_runner/kernel/run_catalog.ex
andreasronge: Fixed in 484ebaf, and you identified the root cause exactly: I was committing to decoded values, and a footer's `artifact_digest` is a stored claim a metadata-only probe never recomputes — so editing `record_count` beside it changed the row while leaving my commitment alone.

Both halves now commit to `{present, descriptor identity, sha256 of the exact bytes probed}` — the trace's head and tail windows, the container's header and footer — which collapsed the two commitment functions into one shared `half_commitment/1`. Malformed and foreign-version probes carry it too, so a half that reached any bytes always commits to them; only a half that never read any (absent, or identity moved) carries neither.

Your reproduction is now a test: rewriting `record_count` in place changes `catalog_digest` and the row together.

---

lib/ptc_runner/kernel/run_catalog_probe.ex
andreasronge: Fixed in 484ebaf, following `Handle.open/2` as suggested: `lstat` the path, open, `fstat` the descriptor and compare before a single byte is read, then `lstat` again after. A mismatch reports `:unstable` rather than committing the bytes.

This required dropping `:raw` from the open — a raw descriptor is process-local and cannot be `fstat`ed — so the probe now uses a file-server descriptor like `Handle` does. It is two `pread`s per file, opened and closed within the probe, so no handles are retained.

I did not add a test for this one. The window is only reachable by winning a race against the probe, and the alternatives were a test-only export on the probe or a `Process.sleep`, neither of which belongs here. What the existing tests do prove is the negative: if the comparison were wrong in the other direction every probe would report `:unstable`, and all 44 pass. Happy to add instrumentation if you would rather have it covered explicitly.

---

lib/ptc_runner/kernel/run_catalog_probe.ex
andreasronge: Fixed in 484ebaf. `current_inspection_row/4` now starts with `Format.decode_header/1` on the header bytes, so once the versions are recognised as this build's, the whole header is validated the way the artifact reader validates it — including the declared size — before any footer field is projected.

`decode_header_versions/1` keeps ignoring the size deliberately: its whole job is to read the two version fields of a container whose layout this build may not know, and a foreign version may legitimately declare a different header size. It is now only the gate that decides *which* validation applies, never the validation itself.

Your reproduction is now a test: a header-size field of 17 isolates the row as `malformed_metadata`.

---

lib/ptc_runner/kernel/run_catalog.ex
andreasronge: Fixed in 484ebaf. Both halves now claim into one space: a trace states its identities in plaintext and a footer states them only as digests, so I hash the plaintext to meet the footer rather than trusting the footer to name itself. Previously the two halves claimed into spaces that could never intersect, which is why the impersonated run stayed admissible.

One thing I want to flag rather than quietly decide. In your reproduction both entries are now isolated, but they report different reasons: A gets `duplicate_run_identity` (its only defect), while B gets `filename_run_mismatch`, which outranks it in my reason ordering. B genuinely has both defects, and the mismatch is the more specific and more actionable one — its bytes are not the run its name claims. Given the stem is verified against the footer hash, a copied artifact essentially always carries a filename mismatch too, so forcing both to `duplicate_run_identity` would mean suppressing a truer reason.

I read the frozen matrix's requirement as "all claimants isolated", which now holds, and the test asserts both the isolation and each specific reason. If you intended the reason itself to be uniform across claimants, say so and I will reorder `duplicate_run_identity` above `filename_run_mismatch`.

---

lib/ptc_runner/kernel/run_catalog.ex
andreasronge: Fixed in 484ebaf. You are right that the partial check was worse than none — it advertised `{:error, :invalid_catalog}` and then broke that contract by raising out of a projection instead.

`valid_trace_probe?/1` now checks the byte count, descriptor identity, commitment and source kind, the head's `run_id`/`trace_id`/`labels`/`timestamp`/`schema_version`, and every tail field type; `valid_inspection_probe?/1` does the same for both its probed and foreign-version shapes. The rule I applied is that anything the classifier later reads is checked here, without exception.

Your `head.timestamp: 42` case plus seven siblings (head schema, all four tail fields, bytes, commitment) are now a table-driven test asserting `{:error, :invalid_catalog}` for each.

---

lib/ptc_runner/kernel/run_catalog_probe.ex
andreasronge: Closed in `5d6921e2`. The probe now exposes deterministic test synchronization immediately before open and immediately after open but before `fstat`. The regression replaces the inventoried path, waits until the replacement descriptor is open, restores the original path, and then resumes. With the descriptor comparison temporarily removed, the row incorrectly becomes `admissible`; with the fix present, it is isolated as `unstable_entry`. No sleep or probabilistic retry is involved. The 101 focused tests and full repository pre-push gate pass.

lib/ptc_runner/kernel/run_catalog.ex
andreasronge: Decision accepted and frozen in [the #1642 contract amendment](https://github.com/andreasronge/ptc_runner/issues/1642#issuecomment-5449287788). Every duplicate claimant must be isolated, while the single primary reason follows the existing precedence: the correctly named claimant reports `duplicate_run_identity`; the copied artifact under the wrong canonical stem reports the more specific `filename_run_mismatch`. The current implementation and regression are therefore the intended contract.

andreasronge:
## Merge-time cleanup completed

- `5d6921e2` adds deterministic coverage for replacement between initial `lstat` and descriptor verification. The counterfactual test publishes the substitution as `admissible` without the `fstat` comparison and isolates it as `unstable_entry` with the fix.
- The #1642 amendment now explicitly freezes `correlation: "unavailable"`, `trace_present: "unreadable"`, and primary-reason precedence for duplicate claimants.
- Both remaining review threads are resolved.
- The PR body now references the current commits and verification results.
- #1642 is an implementation-ready epic; #1681, #1682, and #1683 track steps 2–4.
- Local focused tests (101), `mix precommit`, the full repository pre-push gate, and GitHub CI all pass.

Process note: this follow-up was manually verified by Codex without the independent-review skill, following the instruction on this PR task. The repository-prescribed independent cumulative review is therefore not claimed here, and I have not applied `ready-to-merge` on that basis.

lib/ptc_runner/kernel/run_catalog_probe.ex
andreasronge: Reviewed `5d6921e2` and it is better than what I declined to write — thank you. Two things I had wrong: I treated "no deterministic test" as a property of the race, when it was a property of the seam I was unwilling to add; and I assumed the seam would have to be a test-only export, when a nil-by-default hook validated for arity is exactly the shape `:listing_hook` already established, so it adds no reachable production surface.

The rendezvous is the part I could not see: swapping before `open` and restoring before the closing `lstat` means both path-based stats observe the original while the descriptor is the replacement, so the `fstat` is the only thing standing between that and a committed row. Verifying it by removing the comparison and watching the row go `admissible` is the check that makes it a real regression rather than a passing assertion.

Confirmed on my side at `5d6921e`: 45 catalog tests pass, `mix precommit` clean with duplication `new=0`, `mix dialyzer` 0 errors.

---

lib/ptc_runner/kernel/run_catalog.ex
andreasronge: Understood — isolation is the invariant, the primary reason follows existing precedence, and the current implementation and regression are the intended contract. Nothing to change; resolving.

---

PR #1703, opened 2026-08-28
fix(kernel): name where an explicit failure value went

### LINKED ISSUE #1693: friction(diagnostics): the value passed to (fail ...) is discarded from the envelope, trace and inspection record, so every deliberate failure reports only "the workflow failed"
## Repro

`fail` is the language's explicit failure outcome, and the shipped MCP
documentation teaches it as the way to report a structured error. From
`ptc docs mcp`, the prompt-visible wrapper pattern it recommends:

```clojure
(let [response (tool/workspace.read args)]
  (if (= :ok (get response :status))
    (get response :value)
    (fail response)))
```

So: put a payload in `fail`. Then look for it.

```console
ptc init failtest && cd failtest

python3 -c "
import json;d=json.load(open('ptc-project.json'))
d['artifacts']['inspection']=True; d['artifacts']['result']=True
json.dump(d,open('ptc-project.json','w'),indent=2)"

cat > main.clj <<'EOF'
(ns main)

(defn run [input]
  (fail {"status" "error" "reason" "upstream_rejected"
         "detail" "the vendor returned HTTP 429" "retry_after_s" 30}))
EOF

ptc run ptc-project.json --envelope /tmp/e.json
for t in upstream_rejected retry_after_s 429; do
  echo -n "$t: "; grep -rl "$t" .ptc/ /tmp/e.json 2>/dev/null || echo ABSENT
done
```

## Observed

```console
$ ptc run ptc-project.json --envelope /tmp/e.json
error: execution/workflow_failed: the workflow failed (run_ref: cmd-48bx1ww4x8kbf0nn2ywbxah623)
exit 5

upstream_rejected: ABSENT
retry_after_s:     ABSENT
429:               ABSENT
```

All three artifacts were written and none contains any part of the payload:

```
.ptc/traces/cmd-48bx1ww4x8kbf0nn2ywbxah623.jsonl
.ptc/inspection/cmd-48bx1ww4x8kbf0nn2ywbxah623.ptcins
.ptc/envelopes/cmd-48bx1ww4x8kbf0nn2ywbxah623.json
```

The envelope's error record is empty in every field that could carry it:

```json
{"code":"workflow_failed","message":"the workflow failed",
 "notes":[],"path":null,"span":null,"subject":null,"source":null,
 "phase":"execution","retryable":false,"provider_activity":false}
```

The value the workflow deliberately constructed to explain its own failure is
discarded at the boundary. There is no flag, no artifact and no destination that
retains it.

For contrast, in the same binary and the same release, ordinary evaluator errors
now surface properly — this is the half of #1581 that landed and it works well:

```console
# (defn run [input] (return (/ 1 0)))
error: execution/evaluation_failed: the evaluation failed
evaluation: arithmetic_error: division by zero

# a runaway loop
evaluation: loop_limit_exceeded: Loop iteration limit exceeded (1000 iterations). Use reduce/map over a finite sequence instead, or split work across smaller loops.
```

So the mechanism for attaching evaluation detail to a terminal outcome exists and
is wired. `fail` is the one path that does not use it.

## Expected

The `fail` value reaches at least one artifact — most naturally `error.subject`
or `error.notes` in the envelope, and a trace event. A workflow that fails
deliberately should be able to say why.

## Why it hurts a new user

`(fail …)` is how a PTC-Lisp workflow reports an application-level failure, and
`ptc docs mcp` shows it carrying the whole tool-error envelope. A user who
follows that pattern gets `the workflow failed` and nothing else — the more
carefully they construct the payload, the more information is lost.

It also collapses distinctions a CI script needs. Exit 5 already covers a rejected
credential, a nonexistent model, and the workflow's own `(fail …)`; `error.code`
separates the first two, but every deliberate failure in every workflow shares one
opaque `workflow_failed`. A tester building a CI harness this pass listed exactly
this as one of three failure classes their script could not tell apart.

Two smaller things found alongside, both cheap:

- `(doc "fail")` returns only `(fail ...)` and `nil` — no documentation for a core
  control form.
- `ptc docs ptc-lisp` §5.20 does document it, but in terms of the Elixir API:
  *"Direct `PtcRunner.Lisp.run/2` reports it as a successful evaluation whose
  `Result.return` is `{:__ptc_fail__, value}`."* A binary-only user has no
  `PtcRunner.Lisp.run/2`, and the section never says what the CLI does with the
  value — which, as above, is nothing.

## Severity

**Time-sink** — silent loss of information the user explicitly authored, on a
documented pattern.

## Suggested fix

Publish the `fail` value into the terminal diagnostic the way evaluator errors are
now published (`error.subject`, or a `notes` entry, subject to the existing byte
ceiling and data-class rules). Give `fail` a `(doc …)` entry, and restate
§5.20 in terms of what a `ptc run` caller observes rather than the embedding API.

## Environment

`ptc 0.14.0 (a5343c7d, clean)`, macOS 26.3.1 arm64 (Darwin 25.3.0), packaged
`ptc-0.14.0-macos-arm64.tar.gz`. No provider needed.

Bug. Related to #1581, whose evaluator-error half is confirmed fixed in this
release; `fail` is a distinct path (an explicit language-level outcome, not an
evaluator error) and was not covered by it.

andreasronge:
Closes #1693.

## What the issue got wrong, and what it got right

The report says the value passed to `(fail …)` is discarded from every artifact. It is not: it is written verbatim to the private inspection record as an `explicit-failure-value` and reads back complete through `private-run-analysis-v1`. The repro's `grep -rl` is a false negative — `.ptcins` is a binary-framed container, so GNU grep skips it without `-a`.

The suggested fix (publish the value into `error.subject`/`notes`) is rejected: canonical events and the envelope are payload-free by construction, and a `fail` value is arbitrary application data. The claimed asymmetry with evaluator errors is not one — `arithmetic_error` and friends come from a closed catalog.

What is real is that nothing said where the value went, and in two cases it was genuinely dropped in silence.

## The change

A new `execution/explicit_failure` diagnostic (exit 5) whose message names the retention outcome:

| outcome | message |
|---|---|
| `retained` | its value is retained in the run's private inspection record |
| `unrequested` | … the run published no inspection artifact — set `artifacts.inspection` in the project configuration, or pass `--inspect FILE` |
| `unwritten` | … the run's inspection artifact did not reach its destination |
| `oversized` | … it exceeded the terminal result ceiling |
| `unrepresentable` | … it cannot be represented as JSON |

Each is one closed literal. The value itself still never reaches the envelope or the trace. Retention is settled at the outcome boundary against `artifact_state`, because a sink accepting the record is not the same as the artifact landing — and an execution diagnostic outranks the publication failure that would otherwise be the only sign.

A CI script can now separate a deliberate failure from an infrastructure one on `error.code`. Telling one deliberate failure from another remains what the inspection record is for; the trace's `run-stopped` event already carries the payload-free `failure_kind_fingerprint` for grouping.

Also fills in the empty `fail` and `return` reference entries and restates specification §5.20 for a `ptc run` caller rather than `PtcRunner.Lisp.run/2`.

## Review

Three passes via the coding-agent review workflow, six findings, all fixed:

1. retention claimed before the artifact was published;
2. projection failures misreported as size-limit failures (`(fail (fn [x] x))` blamed the byte ceiling);
3. docs promised unconditional retention;
4. the sourced branch of the row was not pinned to the catalog literal, so the schema admitted a pairing the producer refuses to build;
5. `artifacts.inspection` is not settable on a direct `ptc run <manifest>`;
6. a `fail` re-raising a Kernel refusal keeps that refusal's class, which the docs stated categorically.

## Left open

`execution/model_output_truncated` has the same unpinned-sourced-message hole as (4). It is an unrelated row and a separate slice; not touched here.

## Verification

`mix test` (7477 passed), `mix precommit`, `MIX_ENV=dev mix docs --warnings-as-errors`, and the issue's own repro across all five retention outcomes.

https://claude.ai/code/session_01NRb3mX8QLrzDUTMHbbUp91


andreasronge:
## Rebase Summary

Rebased onto `origin/main` (77126ff8) successfully.

### Conflicts Resolved

**Hand-written source (2 files):**
- `lib/ptc_runner/kernel/command_diagnostic.ex` — both main and the branch added different aliases (`LimitConfigurationDiagnostic` vs `ExplicitFailureDiagnostic`); kept both in alphabetical order
- `lib/ptc_runner/kernel/diagnostic_catalog.ex` — same alias conflict, resolved identically

**Generated files (3 conflicts):**
- `priv/schemas/ptc-command-envelope-v4.schema.json` — took upstream, regenerated
- `site/reference/cli/index.html` — took upstream, regenerated

All generated artifacts regenerated via `mix ptc.gen_docs`. Spec checksums updated for Section 5 changes.

### Commit Structure
All 3 original commits preserved:
1. `fix(kernel): name where an explicit failure value went`
2. `fix(kernel): settle explicit-failure retention against the artifact`
3. `fix(kernel): pin the sourced explicit-failure message and name both remedies`

### Verification
- `mix compile` ✓
- `mix precommit` ✓ (duplication baseline=36, current=36, new=0)
- `MIX_ENV=dev mix docs --warnings-as-errors` ✓
- All CI checks passed ✓

PR is now mergeable (`mergeable_state: clean`).

---

PR #1711, opened 2026-08-29
fix(limits): reject event payload budgets that cannot finalize a run

### LINKED ISSUE #1708: bug(limits): reject event payload budgets that cannot finalize a run before execution
## Summary

#1605 closed the `normal_event_count` hole and, in doing so, opened the same hole on `event_payload_bytes`.

PR #1707 raised the `event_payload_bytes` catalog minimum from 1 to 4,826 and published `4,826–2,592,000,000` as the accepted range. That minimum does not run. Values `4,826` through `6,767` pass `validate` and `doctor` clean and then collapse to exit 70 `internal/internal_error` without a trace — the exact failure shape #1605 was filed to eliminate, now sitting inside the range that fix authored.

The derived floor omitted the `usage` map that every `run-stopped` payload carries.

## Reproduction

Verified on `origin/main` at `77126ff8` (the #1707 merge), stock `ptc init` project — zero providers, zero capabilities, zero missions.

~~~console
ptc init n1605b
cd n1605b

# ptc.json: "limits": {"event_payload_bytes": 4826}   <- the documented minimum
ptc validate ptc-project.json   # exit 0
ptc doctor   ptc-project.json   # exit 0, all checks pass
ptc run      ptc-project.json   # exit 70
~~~

~~~
payload=4826  validate=0  doctor=0  run=70  internal/internal_error: the command failed internally
payload=5000  validate=0             run=70  internal/internal_error: the command failed internally
payload=10000 validate=0             run=0   {"greeting":"hello world"}
~~~

Bisected boundary on that project: **6,767 fails, 6,768 passes**. The dead window is 4,826–6,767, and its lower bound is the catalog minimum #1707 introduced.

Because the stock project declares no capabilities and no missions, this window is entirely application-independent. Nothing about the manifest's content is involved.

## Root cause

Normal trace admission has **two** capacity gates. #1605 scoped its structural contract to the first one and satisfied its acceptance criterion literally — "the catalog minimum for `event_payload_bytes` is admitted by normal sink preflight". It is. The gate that fails is the second one.

1. `EventSink.prepare/3` → `terminal_payload_capacity?/3` → `EventBudget.normal_terminal_payload_capacity?/1`, which is `limit >= 4_826`. Admits.
2. `RunConfig.new/1` → `EventSink.terminal_usage_capacity?/3`, called with `maximum_terminal_usage(workflow, missions, limits)`. Refuses.

`EventBudget.minimum_normal_payload_bytes/0` was derived from the saturated `events-dropped` map plus the maximum bounded terminal reason. But `terminal_usage_capacity?/3` sizes the complete `run-stopped` payload, which is `%{outcome:, reason:, usage:}` — and `usage` was never part of the derivation. Measured on `origin/main`:

~~~
terminal_usage_capacity?(payload=4826, usage=%{})              => true
terminal_usage_capacity?(payload=4826, usage=%{closed?: true}) => false
~~~

One boolean key overflows the floor. Every real run carries far more: `maximum_terminal_usage/3` always sets `closed?`, `remaining_ms`, `subordinate_evaluations`, `subordinate_source_checks`, `protocol_errors`, `evaluation_memory_bytes`, `evaluation_history_bytes`, `evaluation_continuation_bytes`, `evaluation_busy?`, `errors`, `capability_refusals`, and `llm_budget`.

When that gate returns false, `RunConfig.new/1` falls through its `with` to the catch-all `_ -> {:error, :invalid_run_config}`. `RunBuilder.failed_build/4` propagates it and the command layer has no row for it, so it lands on `{:internal, :internal_error, 70, false, "the command failed internally"}`.

## Structural contract

The requirement decomposes into a fixed part and an application-scaled part. They need different treatments.

**Fixed part — application-independent, belongs in the catalog.**

Every field of `maximum_terminal_usage/3` except the capability and mission maps is present in every run and bounded by catalog maxima. `EventBudget.minimum_normal_payload_bytes/0` must be derived from the complete maximum bounded `run-stopped` payload:

~~~text
minimum_normal_payload_bytes =
  retained size of %{
    outcome: :error,
    reason: <maximum bounded terminal reason>,
    usage: <maximum bounded fixed usage fields at catalog maxima,
            with an empty capability inventory>
      |> Map.put(:events_dropped, <saturated reachable drop map>)
  }
~~~

evaluated with every integer field at its catalog maximum, so the floor holds for any admitted limit set. This alone closes the reproduction, because the stock project's application-scaled part is empty.

**Application-scaled part — cannot be a catalog minimum.**

`maximum_call_map/3` keys the workflow and mission call maps by declared capability name, and `evaluations_by_mission` is keyed by mission name. The requirement therefore grows with the manifest and with resolved provider tool names. No static minimum can bound it, and it must not be one.

## Where the scaled check can run

Not in `validate`. The capability inventory is `providers.workflow.capabilities` and each mission's `provider_occurrences`, assembled in `RunBuilder` after provider assembly; an MCP-backed provider's tool names are only known after acquisition. Do not pretend otherwise, and do not add a provider-inert approximation that is wrong in either direction.

The requirement is that the failure be **classified** correctly wherever it is first computable, not that it move earlier. `RunConfig.new/1` must return a distinguishable reason instead of the generic `:invalid_run_config`, and that reason must reach the command layer as a configuration diagnostic at exit 3 with no trace requirement — never `internal/internal_error`.

## Accepted diagnostics

Static values below the raised catalog minimum use the existing schema diagnostics, as in #1605:

- manifest: `application/schema_violation` at `/limits/event_payload_bytes`, naming the minimum rule;
- host: `host/host_schema_invalid` at `/limits/event_payload_bytes`, naming the minimum rule.

The application-scaled refusal uses a new V4 diagnostic:

~~~text
application/limit_capacity_invalid
~~~

with a closed bounded message of this form:

~~~text
event_payload_bytes effective limit P is below the required R bytes for this application's resolved terminal usage; raise limits.event_payload_bytes, and its installed host ceiling if it is lower, or declare fewer capabilities
~~~

`P` and `R` are catalog-validated integers using the bounded integer patterns already established by `LimitConfigurationDiagnostic`.

Use a new code rather than a second message form under `application/limit_configuration_invalid`. That code's message is a closed three-integer pattern whose `valid_message?/1` round-trips through `LimitConfiguration.required_normal_event_bytes/1`; a second form would force the schema pattern into an alternation and the validator into dispatch, for a different relationship with a different remedy.

## Decisions

- Derive the payload floor from the complete maximum bounded `run-stopped` payload, `usage` included, not from the drop map alone.
- Keep the floor application-independent and evaluated at catalog maxima, so it holds for every admitted limit set.
- Do not attempt the application-scaled check in `validate` or `doctor`; the inventory is not resolved there.
- Do not classify a capacity refusal as `internal/internal_error`. It is a configuration failure, exit 3, no provider activity beyond whatever the inventory already required, no trace required.
- Remove the generic `{:error, :invalid_run_config}` fallthrough for this case specifically; leave the catch-all in place for the reasons that genuinely are internal.
- Derive the floor constant by measurement in the test suite. The current `limit_catalog_test` asserts `minimum == 4_826` against the module attribute it is meant to police, and `event_sink_test`'s search for the smallest admitted payload resolves through the same constant. Both are circular. The load-bearing check is a real saturated `run-stopped` finalization; make that the authority.

## Acceptance criteria

- The stock `ptc init` project runs at every documented-legal `event_payload_bytes`, including the catalog minimum.
- One byte below the new catalog minimum fails manifest and host schema validation with a path to `/limits/event_payload_bytes`.
- Host values below the new minimum fail host schema validation.
- A manifest whose resolved capability inventory pushes the requirement above its payload limit fails with `application/limit_capacity_invalid` at exit 3, reporting `P` and `R`.
- `RunConfig.new/1` no longer returns `:invalid_run_config` for a terminal-usage capacity refusal.
- No limits-only edit to a manifest can reach exit 70 `internal/internal_error`.
- At the exact derived boundary, a saturated `run-stopped` payload finalizes; one byte less is refused by schema validation.
- The floor constant is proven by a test that measures a real saturated `run-stopped` payload, not by asserting the constant against itself.
- Limit catalog tests, generated schemas, the generated limits reference, `docs/maintainers/kernel.md`, command fixtures, and integration tests are updated.

## Ordering

**Unblocked.** #1707 is merged at `77126ff8`. The new diagnostic targets the current V4 command envelope.

## Severity

Bug, and a regression in scope: #1707 moved this failure into the documented accepted range. Same class as #1605 — a documented-legal value passes `validate` and `doctor` and then exits 70 with the one message a caller cannot act on.

## Related

Follow-up to #1605 (PR #1707).

andreasronge:
## Summary

- Raise the `event_payload_bytes` catalog floor to **8,211** so a complete maximum `run-stopped` payload — drop map, terminal reason, and usage at catalog maxima with empty capability/mission inventory — is admitted. Stock `ptc init` projects now run at every documented-legal payload, including the minimum.
- Classify application-scaled terminal-usage refusals as `application/limit_capacity_invalid` (exit 3, no trace required) instead of collapsing through `:invalid_run_config` to exit 70 `internal/internal_error`.
- `RunConfig.new/1` returns `{:error, {:limit_capacity_invalid, p, r}}`. Both the provider-free owner path (`RunBuilder.environment_failure_diagnostic/3`) and the active-provider marked-failure classifier project that reason so `provider_activity` stays evidence-correct.
- The closed remedy names both levers: raise `limits.event_payload_bytes` (and its host ceiling if lower), **or declare fewer capabilities or missions**.
- Doctor/validate still skip the scaled check. Values below the new catalog minimum stay `application/schema_violation` / `host/host_schema_invalid`. `limit_capacity_invalid` is excluded from doctor application rows and their envelope schema.

Closes #1708.

## Test plan

- [x] Stock project at catalog minimum: `validate`/`doctor`/`run` succeed
- [x] One byte below the minimum: manifest and host schema diagnostics at `/limits/event_payload_bytes`
- [x] Fat mission names at catalog min: `run` exits 3 with `application/limit_capacity_invalid`; `validate`/`doctor` still pass
- [x] Provider-backed lifecycle: same refusal after acquisition with `provider_activity: true`
- [x] Saturated `run-stopped` finalization at the floor (`measured <= minimum`)
- [x] Local pre-push gates green (docs, 7504 tests, Dialyzer, release, Viewer)

Made with [Cursor](https://cursor.com)


andreasronge:
Superseded by #1714, which lands the same 8,211 floor and `application/limit_capacity_invalid` with a regression guard that derives the maximum terminal usage key set from a real `RunState.usage/1` (`core_contract_test.exs`), and derives test payload limits from `EventBudget.minimum_normal_payload_bytes/0` instead of hardcoding them.

PR #1977, opened 2026-09-15
docs(kernel): clarify private serving refusal

### LINKED ISSUE #1944: Serving: decide and add authorized private in-memory publication
## Problem and reproduction

Follow-up from merged PR #1942's retrospective and the maintainer decision in #1939. A provider-free serving manifest with `events.policy: private` is rejected by `ServingTemplate.from_directory/3` with `{:error, :private_result_unservable}`. The regression `private templates are rejected before serving` in `test/ptc_runner/kernel/serving_template_test.exs` records this intentional release restriction.

## Decision required

Private serving was explicitly deferred from #1918. Decide whether to authorize a private in-memory destination owned by the gateway request, or retain construction-time refusal. Endpoint possession alone must not implicitly override the private event policy.

If authorization is selected, specify who can obtain the publication capability, its binding to the calling worker and sealed outcome, single-use consumption and close semantics, and absent/invalid authority behavior before implementation. Define closed safe outcome fields and precedence against cancellation, deadline, publication and cleanup failures.

## Scope after the decision

- Add the authorized in-memory serving destination and public API only under the selected disclosure contract.
- Preserve ordinary command/artifact withholding and prohibit private details in safe error metadata.
- Retain admission capacity through calling-worker publication and cleanup; uncertain cleanup wins and fences admission.
- Add public constructor-to-call coverage for authorized private success, absent/invalid authority, cancellation and publication/cleanup races. Update the owning public documentation.

This is an ordinary follow-up outside the completed #1918 collection, not a prerequisite for its closed members. Related: #1465. Do not add a private publication bypass to the existing serving API before the decision.

andreasronge:
## Summary

- Record that private serving remains refused at template construction because the gateway has no authorized private destination.
- Clarify that endpoint possession and `allow_write` do not override private event policy.
- Document the deliberate normal-policy change and content re-pin path, and regenerate the published gateway reference.

Closes #1944

## Validation

- `mix test test/ptc_runner/kernel/serving_template_test.exs`

## Retrospective

- Untracked follow-up work: none.
- Missing or incorrect repository instruction: none.


andreasronge:
Every claim in #1977 is now in #1978 or already on main:

